PR #7, opened 2017-02-25
Fixes for gdscript syntax checking

Koivulehto:
Hi. I noticed that gdscript syntax checking had a couple of issues. Namely that the comments weren't coloured and that comments had their syntax checked. While working I also saw that indentation checking couldn't deal with empty lines. To fix these issues I changed the code so that comments are not parsed along with the rest of the code and nextline checks until it finds a line containing text. In addition I fixed a couple of typos and deprecations. 

To demonstrate the issues at hand, I created a test_files folder and inside it test.gd and test2.gd. In the current release, test2.gd has complaints on 3 lines out of 4, even though it contains only comments.

You have done great work on the extension and I hope this helps to make it even better.


Geequlim:
Wow that seems wonderful !

src/gdscript/symbolparser.ts
```diff
@@ -33,7 +33,7 @@ class GDScriptSymbolParser {
         documents: {},
         constvalues: {}
     }
-    const text  = content;
+    const text  = content.replace(new RegExp(/#.*$/, "gm"), ""); //excludes comments from being checked for syntax
```
Geequlim: This kills documentation on top of symbols.

Koivulehto: I didn't catch that bug when testing. I will fix it asap.

Koivulehto:
Reverted changes in symbolparser.ts. After reverting changes I was unable to fix variables in comments getting warnings saying they are unused. Rest of the fixes should still work.

PR #11, opened 2017-03-15
Remove notification when there is no error. Fixes #8

### LINKED ISSUE #8: "Update GDScript symbols done" notification on program open intrusive
"Update GDScript symbols done" notification on program open is intrusive. Can it be done in a less intrusive manner, hidden, or have an option to be hidden?

tapir:


Geequlim:
The message is moved to status bar so The code cannot be merge.
And also close #8

Issue #15, opened 2017-03-31
line continuation does not get recognized


shawniac:
The following is *valid* GDScript, at least it works in the engine itself with no complaints, but gets highlighted as an error. (see **Line 64** and **Line 71**)

Screenshot:
![31_03_2017](https://cloud.githubusercontent.com/assets/1819638/24541343/5abfbd50-15f7-11e7-84f5-b2ec83ce4543.png)

Geequlim:
I have no idea about this :(

Geequlim:
Fixed.
Thanks

PR #16, opened 2017-04-28
Fixed two issues within validateExpression (multiline comments and multiline statements)

arrkiin:
While browsing thru some Godot projects I discovered some issues with the validation of some correct GDScript statements. Because I had some trouble debugging the extension I fixed the `launch.json` file.

### Multiline comments
**Problem:** Within multiline comments the keywords like `if`, `for` etc. were treated as if they are normal code. So they produced validation errors like `':' expected at end of the line.`.
**Solution:* Now the beginning and the end of these regions are recognized and are not processed by the following keyword checks.

### Multiline statements
**Problem:** Multiline statements like described in issue #15 were not handled correct.
**Solution:** The beginning of a multiline statements, indicated by an ending backslash `if a = 10 and \` is recognized and processed until the statement ends with a colon `b = 20 :`. A missing backslash within a multiline statement of more than two lines is also recognized and leads to an `': or \' expected at end of the line.`

### Debugging configuration
`"outFiles": ["${workspaceRoot}/out/src/**/*.js"],`

### DIFF
--- a/src/gdscript/diagnostic.ts
+++ b/src/gdscript/diagnostic.ts
@@ -77,6 +77,8 @@ class GDScriptDiagnosticSeverity {
 
   private validateExpression(doc: vscode.TextDocument) {
     let diagnostics = [];
+    let expectEndOfLine = false;
+    let commentActive = false;
     const text = doc.getText();
     const lines = text.split(/\r?\n/);
     lines.map((line:string, i: number) =>{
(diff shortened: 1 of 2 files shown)


src/gdscript/diagnostic.ts
```diff
@@ -85,18 +87,36 @@ class GDScriptDiagnosticSeverity {
       if(matchstart)
         curLineStartAt = matchstart.index;
       // ignore comments
-      if(line.match(/^\s*#.*/) || line.match(/^#.*/)) return
+      if(line.match(/^\t*?""".*?"""/))
```
Geequlim: """This is not a comment in GDScript""" it's a string.

arrkiin:
I misinterpreted the multiline string colorization and behaviour in the godot editor. I splitted the multiline statement extension into a seperate pull request #20. Therefore I close this request.

PR #21, opened 2017-05-02
Multiline statements

arrkiin:
Because of a wrong interpretation of multiline strings I splitted the original pull request #16 into this concrete pull request for multiline statements. Pull request #16 will be closed.

### Multiline statements
**Problem:** Multiline statements like described in issue #15 were not handled correct.
**Solution:** The beginning of a multiline statements, indicated by an ending backslash `if a = 10 and \` is recognized and processed until the statement ends with a colon `b = 20 :`. A missing backslash within a multiline statement of more than two lines is also recognized and leads to an `': or \' expected at end of the line.`


Geequlim:
Merged thanks !

PR #37, opened 2017-08-15
Add project root configuration settings

kosz78:
In some cases godot project can be in the separate folder of main project, I have added configuration parameter to handle this case.


Geequlim:
Thank you.
It may break the commands to run the game/scene

kosz78:
I have added support for run scene, project

PR #56, opened 2018-03-18
Added handling of paths with spaces in them

RivenCode:
For the editor path, we use PowerShell syntax, when required.

### DIFF
--- a/src/tool_manager.ts
+++ b/src/tool_manager.ts
@@ -183,7 +183,7 @@ class ToolManager {
       let pathFlag = "-path";
       if (vscode.workspace.getConfiguration("GodotTools").get("godotVersion", 2.1) >= 3)
         pathFlag = "--path";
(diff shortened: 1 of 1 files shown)


src/tool_manager.ts
```diff
@@ -259,6 +260,20 @@ class ToolManager {
...
 
+  private escapeCmd(cmd: string) {
+      // Double quote command (should work in at least cmd.exe and bash)
+      let cmdEsc = `"${cmd}"`;
+
+      // Fetch Windows shell type
+      let shell = vscode.workspace.getConfiguration("terminal.integrated.shell").get("windows", "");
+
+      // For powershell we prepend an & to prevent the command being treated as a string
+      if (shell.endsWith("powershell.exe") && process.platform === "win32") {
```
Geequlim: Does 64bit of the VSCode is also valid for `win32` ?

RivenCode: Yep, Node.js returns `win32` regardless.

Issue #25, opened 2017-05-24
multi-line if statement showing as error


JimmyJames707:
Hi,
This is legal syntax in Godot, but is reported as an error in your extension.

![screenshot_2017-05-25_04-41-20](https://cloud.githubusercontent.com/assets/6817086/26420017/d76ff30a-4104-11e7-9e6e-cd56a211bb38.png)

Geequlim:
The new version fixed the problem and will be published when godot 3.2 RC release.

Geequlim:
I'm closing this as the new version has fixed this issue.
You can try the new version of the plugin with godot 3.2 from the [release page](https://github.com/godotengine/godot-vscode-plugin/releases)

PR #79, opened 2018-09-03
single line if-else statement

akiraKido:
This fixes the issue where single line if-else statements are diagnosed as an error:
```gdscript
variable = true if x == y else false
```
The regex(`.*?if.*else.*`) might be too broad though...

### DIFF
--- a/src/gdscript/diagnostic.ts
+++ b/src/gdscript/diagnostic.ts
@@ -112,11 +112,8 @@ class GDScriptDiagnosticSeverity {
           return
         if(line.match(new RegExp(`.*?#.*?\\s${keywords[1]}\\s.*?`)))
           return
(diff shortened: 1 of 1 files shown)


fractile81:
Agreed that the regex is a bit too broad.  The regexp before it (line 115) looks like it should cover your example - see https://regex101.com/r/PqK3de/1 - and copy/pasting that expression into a .gd file didn't throw an error.  Can you explain how to reproduce this error?

If this is in reference to #25 , this wouldn't address that since it's for a multi-line if-statement.

akiraKido:
You were absolutely right. The regexp before should've covered it, but my line was actually as the following:
```gdscript
is_on_tree = true if "Tree" in current_bottom.name else false
```
(and yes, I do know that it could've been just `is_on_tree = "Tree" in current_buttom.name`)
The `"` wasn't caught in there, so I've fixed the previous regexp to accept double quotes as well.

PR #84, opened 2018-10-10
Close existing terminal when executing "run" commands

lleaff:
I often find myself with multiple game windows when running projects using this extension.
I prefer Godot's editor behavior of closing the existing game windows before launching another instance.

This PR adds a simple check followed by `Terminal#dispose` if a terminal found with the same name as our extension ones is found. Terminals are renamed to a more specific "GodotTools" to avoid a very hypothetical conflict with another extension.

Cheers,


Geequlim:
Thank you !

PR #102, opened 2019-05-30
minor addition in configuration and grammar files

ankitpriyarup:
Added following things in grammar & configuration:
Folding
image

Added new keywords such as: switch, preload, assert, etc matching with current version of gdscript documentation

"//" style comment at the top of snippets.json was causing a problem hence removed it (it's redundant)

### DIFF
--- a/configurations/GDScript.tmLanguage.json
+++ b/configurations/GDScript.tmLanguage.json
@@ -91,7 +91,7 @@
     },
 
     "keywords": {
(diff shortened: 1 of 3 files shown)


Geequlim:
`switch` and `case` are not keyworks of GDScript.

Here are the keywords in GDScript
https://github.com/godotengine/godot/blob/defd9602764c13d1b7753cf879bdbcdf56335b33/modules/gdscript/gdscript.cpp#L1774-L1835

And also we should follow this for builtin classes
https://github.com/godotengine/godot/blob/bc7178422e98c86c7c8c9d9e572a9c6941836d79/core/variant.h#L82-L121

You can update your PR with commit  `--force`

PR #106, opened 2019-07-24
Add `class_name` keyword to syntax highlighting

Calinou:
This adds the `class_name` keyword that was added to Godot 3.1.

PS: What's the difference between `GDScript.full.tmLanguage.json` and `GDScript.tmLanguage.json`? Are these used in different contexts?


Geequlim:
The `GDScript.full.tmLanguage.json` is not use now.

Thanks.

PR #108, opened 2019-08-03
Add fallback for shell var

acklavidian:
Plugin doesn't seem to work on any of my linux machines. This shell var seems to be null on linux. Seems like it might be the case on MacOS too. When its null the console reports a lack of `endsWith` method on the null var. I am adding a fallback for the shell var that is an empty string. I am not sure if this is the best way to fix. But it does the trick for me.


Geequlim:
Thanks

PR #112, opened 2019-08-30
Language Server Markdown Documentation

ankitpriyarup:
This PR provides support for the markdown documentation by first parsing the symbol by - either relying on textDocument/definition provided by LSP (to resolve ctrl + click) or by relying on a simple regex parser to parse the text from cursor position. Symbol is then looked through the dumped index file from doc and later appropriate markdown documentation is show in the side panel.

Refer following Godot Source as server: https://github.com/ankitpriyarup/godot/tree/lsp


Geequlim:
The native documentation is now implemented with custom specifications. Documentations are send from the server realtime and renderer as in a spatial webview in VSCode. So we don't need to dump and load to markdown files.

PR #118, opened 2019-10-03
fix string literal highlighting

### LINKED ISSUE #88: Sequence "\\" gets wrongly escaped
The `'\'` char in String is not correctly escaped and the editor interprets everything below this expression as I think String sequence:

peek 2019-01-18 20-43

Greetings

pduzinki:
fixes #88


Geequlim:
Thanks!

PR #134, opened 2019-12-29
Fix syntax highlighting for camelCase identifiers

Bromeon:
Currently, any occurrences of PascalCase identifiers (even as parts of other words) are recognized as classes.

This assumes snake_case convention for all methods and variables and makes it impossible to use camelCase. While this is the recommended GDScript style, the syntax highlighter should allow for different styles as long as it can do so unambiguously. This is already done for existing rules, but overridden by one rule with an overly general regex pattern.

This commit modifies the catch-all rule for the `parscal_class` group to only capture whole words.
For clarity, it renames `parscal_class` to `pascal_case_class`.

Before:
before

After:
after

### DIFF
--- a/configurations/GDScript.tmLanguage.json
+++ b/configurations/GDScript.tmLanguage.json
@@ -27,7 +27,7 @@
     { "include": "#any-method" },
     { "include": "#any-property" },
     { "include": "#extends" },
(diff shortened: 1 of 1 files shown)


Geequlim:
Thanks

PR #141, opened 2020-02-09
Switched this plugin from Websockets to the raw TCP implementation

ofrank123:
I've switched this plugin from using the websocket implementation to the newer raw TCP server on [this pull request](https://github.com/godotengine/godot/pull/35864), as requested by @Geequlim.


ofrank123:
This can be merged I believe, the PR on the main Godot repo was just merged.

Geequlim:
We'd better keep the websocket support and add a configuration to let user select the protocal(by godot version). So we can release new version without break the 3.2 support in the future.

ofrank123:
I think @akien-mga was going to cherry pick the commit for 3.2 anyways? So it shouldn't be a huge problem.

akien-mga:
Well 3.2 is released already. I'll cherrypick the changes, but probably for 3.2.2, since 3.2.1 has already had a Release Candidate and I'll likely publish it almost as is in coming days.

So it might make sense to use WebSockets for 3.2 - 3.2.1, and TCP for 3.2.2+.

ofrank123:
Okay, I'll work on making it compatible with both then.

Geequlim:
@ofrank123  Any update about this ?

TicklesTheBrain:
An updated version of plugin that works with 3.2.2 would be useful for testing 3.2.2 beta.

ofrank123:
Haven't been free for a bit due to covid, but I'm back working on this.  I'll try to sort these issues out ASAP.  Sorry for the wait!

ofrank123:
This is now an option in the config, which is off by default.  This should make it so users not using the new beta will experience no change.  I recommend that it is set to on by default when 3.2.2 is released.

PR #146, opened 2020-02-16
Add debugging support to the extension

Razoric480:
I believe I've gotten the debugger fairly well integrated with the extension - though it kind of floats apart in the VSCode context, so there isn't much collaboration required with the existing systems. It doesn't conflict with it, though, so I guess that's the important part.

## Supports ##

- Step-in/next
- Step
- Continue
- Pause (though since Godot pausing doesn't report a breakpoint, the editor doesn't change, but hitting pause again will resume)
- Locals, members and globals scopes for variables while broken
- Scope drilling and objects-within-objects, including for Object types.
- Exceptions break differently and do not allow continue/step-in/step
- Stack trace and jumping between those frames
- Terminate
- A boolean in the launch config permits control over whether to launch a Godot instance, or wait for the remote debugger to launch (I.E. from the editor)

~~## Missing ##~~

~~Currently, things that extend Object (KinematicBody2D, Camera2D, etc) will show their class type, but their properties are not yet exposed to be printed. Given the Godot debugger is _very_ liberal about how much information it gives you when you call for an inspect_object command, this may be a bit excessive to show everything. Still, we do _receive_ the properties, just haven't done anything with them yet.~~

## To be done ##

Set up a branch in preparation for the upcoming [refactor](https://github.com/godotengine/godot/pull/36244) - I haven't looked at the code yet, but from reading the PR blurb it seems that the format of the commands will change somewhat.

---

I'll admit that my Typescript/javascript isn't the best, so any reviews are welcome. But it feels pretty thorough, and I can't think of anything else that's missing.

### DIFF
--- a/src/debugger/communication/server_controller.ts
+++ b/src/debugger/communication/server_controller.ts
@@ -227,6 +227,18 @@ export class ServerController {
 			this.connection = connection;
 			this.godot_commands?.set_connection(connection);
 
+			if (!launch_game_instance) {
+				this.breakpoints.forEach(bp => {
+					let path_to = path
+						.relative(this.project_path, bp.file)
+						.replace(/\\/g, "/");
+					this.godot_commands?.send_set_breakpoint_command(
+						`res://${path_to}`,
+						bp.line
+					);
+				});
+			}
+
 			connection.on("data", buffer => {
 				if (!this.parser || !this.builder) {
 					return;
--- a/src/debugger/communication/server_controller.ts
+++ b/src/debugger/communication/server_controller.ts
@@ -1,0 +1,393 @@
+const TERMINATE = require("terminate");
+import { EventEmitter } from "events";
+import net = require("net");
+import cp = require("child_process");
+import path = require("path");
+import { VariantParser } from "../variant_parser";
+import { Command } from "./command";
+import vscode = require("vscode");
+import { GodotCommands } from "./godot_commands";
+import { CommandBuilder } from "./command_builder";
+import { GodotBreakpoint, GodotStackFrame } from "../godot_debug_runtime";
+import utils = require("../../utils");
+
+export class ServerController {
+	private breakpoints: { file: string; line: number }[] = [];
+	private builder: CommandBuilder | undefined;
+	private connection: net.Socket | undefined;
+	private emitter: EventEmitter;
+	private exception = "";
+	private godot_commands: GodotCommands | undefined;
+	private godot_pid: number | undefined;
+	private inspected_callbacks: ((
+		class_name: string,
+		properties: any[]
+	) => void)[] = [];
+	private last_frame:
+		| { line: number; file: string; function: string }
+		| undefined;
+	private output_channel: vscode.OutputChannel | undefined;
+	private parser: VariantParser | undefined;
+	private project_path: string;
+	private scope_callbacks: ((
+		stack_level: number,
+		stack_files: string[],
+		scopes: {
+			locals: { name: string; value: any }[];
+			members: { name: string; value: any }[];
+			globals: { name: string; value: any }[];
+		}
+	) => void)[] = [];
+	private server: net.Server | undefined;
+	private stack_count = 0;
+	private stack_files: string[] = [];
+	private stack_level = 0;
+	private stepping_out = false;
+
+	constructor(
+		event_emitter: EventEmitter,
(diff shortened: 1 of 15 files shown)


Razoric480:
I changed my mind and added scope drilling for object types on a sudden eureka moment.

Geequlim:
Thanks for your contribution.
You'd better follow the code style. The code style is basicly following the style of godot engine project.

> Using tabs instead of spaces
> Using snake_case for file names
> Using snake_case for member variables and local variables
> Using UPPER_CASE for constants and enumeration
> Using PascalCase for class name

You can install the EditorConfig plugin for you IDE to make it easier.

Razoric480:
Okay, I'll amend it appropriately.

Geequlim:
I can my project but client freezed when a breakpoint enter.

Got error log from the debug console
```
Unsupported command: message:click_ctrl
CommandBuilder.js:36
Unsupported command: 2
CommandBuilder.js:36
Unsupported command: [object Object]
CommandBuilder.js:36
Unsupported command: LinkButton
```

Razoric480:
That should be fixed now - commands/messages that are not supported (yet) will be consumed and skipped, instead of festering and causing bugs/crashes.

Geequlim:
@Razoric480 I'm trying it with a real project. But got the problem with multi-threading

https://github.com/godotengine/godot/issues/2446
https://github.com/godotengine/godot/issues/19809

@reduz  @Faless Any ideal about this?

Geequlim:
I'll compile the multi-threading related stuff with a C++ mudole and test later.

src/debugger/communication/server_controller.ts
```diff
@@ -0,0 +1,332 @@
...
+			connection.on("drain", () => {
+				connection.resume();
+				this.godot_commands?.set_can_write(true);
+			});
+		});
+
+		this.server?.listen(port, address);
+
+		if (launch_game_instance) {
+			let executable_line = `godot --path ${project_path} --remote-debug ${address}:${port}`;
```
Geequlim: We'd better using the godot editor of the plugin configuration instead of on the path.
For example I installed offecial version on the system path as 'godot'. But I most case I'm going to launch the binary compiled with my modules.

Geequlim: We'd better using the godot as the plugin configuration `godot_tools.editor_path`.
And you can run the godot with `GodotTools.run_editor()` method.

Geequlim:
I can launch my project and toggle breakpoint in VSCode now.
But there are something not working as expected.
* I can do step-into but step-over and step-out don't work
* The stack tree doesn't update when stack changed
* The stack tree may be empty when breakpoint triggered
* I can't expand any array dictionary or object in the stack tree.

Razoric480:
### Fixes ###

- godot_tools.editor_path configuration is used by default, but if it's somehow missing, PATH is used  as a fallback.
- Step-out is not supported by Godot officially, which is why I originally skipped it, but I've built it into the adapter by keeping track of amount of stack frames.
- Fixed step-over sending a step-in signal. I blame that everyone uses a different name for it.
- Dictionaries are now parsed properly in scope
- Regression caused arrays to break. They are now fixed.
- Scope now updates when switching frames in callstack.

### Cannot fix ###

The remote debugger does not seem to send any Thread information whatsoever, so I cannot fix this for the extension =(

---

Please give this another test, and see if any of the other remaining bugs still crop up. I am unable to reproduce not being able to expand Object types, so I'd be curious if you'd have a reproducible script/project that does this, if the problem still occurs.

Geequlim:
@Razoric480 Tested your latest commit still can't update the stack tree in some situation. And I can't stop the lauched process after stop debugging.
Here is an example project:
Try toggle a breakpoint at `router.gd:103` and switch pages server time.
[DebuggerTest.zip](https://github.com/godotengine/godot-vscode-plugin/files/4223655/DebuggerTest.zip)

Razoric480:
I'm unable to recreate the issue - works fine on this end. The variable scope updates and the breakpoint triggers, and stopping the debugging stops the launched process.

There may be a conflict with Linux that isn't present on windows? I'm installing Ubuntu onto Hyper-V and will try to test that way.

Razoric480:
No, everything is working on Ubuntu as well :S So I'm a bit at a loss.

PR #148, opened 2020-02-17
Issue 144

bitwes:
Adds `godot-tool.run_godot` command.  This is a non-palette command to be consumed by other extensions.

In testing the command I discovered that I had to add the `terminalName` parameter to prevent my launching of Godot to take over the Godot editor terminal.  I took the liberty of adding two terminal names, one for the editor and one for running the project.  I noticed that running the project would kill the editor.  I wasn't sure if this was "as designed" or not but it appeared to be a bug.  With these changes the running of the project won't kill the editor's terminal window.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ test
 configurations/tmp.txt
 configurations/test.py
 .vscode-test
+package-lock.json
--- a/README.md
+++ b/README.md
@@ -66,14 +66,15 @@ vscode.commands.executeCommand('extension.command', p1, p2, p3 ...)
 
 #### Commands:
 
-__`godot-tools.run_godot(params: string = '')`__ <br/>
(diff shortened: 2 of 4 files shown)


bitwes:
In thinking about this more, maybe the command should just return you the string to launch godot for the current workspace.  That would take terminal management out of the equation and might be more maintainable going forward.  It is conceivable that someone may not want to kill the existing window when running again, and just reuse it.

bitwes:
Let me know you want to address this.  I see the "bug" label so you might agree about the terminal naming issue.  I can keep this PR how it is, or create a new PR that addresses the bug  and another PR to make a command that just returns the run command so that the reusable config and command generation isn't as tightly coupled with the terminal.

bitwes:
Closing this one for now.  I've created a more complicated fix for this that I'll try to get merged instead.

PR #149, opened 2020-02-20
Issue 144

bitwes:
Adds `godot-tool.get_run_workspace_command` non-pallette command that returns a platform dependent string to launch Godot for the currently opened workspace.

Adds `godot-tool.run_godot` command. This is a non-palette command to be consumed by other extensions.

In order to break out the logic for creating the command string I had to refactor the `run_editor` method to make the code accessible.  As I was doing this, I removed the dialog to find the Godot Editor and replaced it with an error message.  I did this for two reasons, one it made making the command easier.  The other was that the dialog was confusing to me when I started using the pluign.   I didn't see anything that told me what I was trying to find.  When you click the "open in editor" button and you get a file dialog you think it wants you to pick a file to open in the editor, not the Godot runtime.

I also noticed that running the project would kill the editor. I wasn't sure if this was "as designed" or not but it appeared to be a bug. With these changes the running of the project won't kill the editor's terminal window.  There is now two terminal names used, one for the editor and one for running the project.

I've verified these changes on Mac and PC.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ test
 configurations/tmp.txt
 configurations/test.py
 .vscode-test
+package-lock.json
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ You can use the following settings to configure Godot Tools:
 - `check_status` - Check the GDScript language server connection status.
 
 #### Use Godot Tools in your VSCode Extension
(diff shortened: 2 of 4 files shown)


bitwes:
Found my way back here after making some changes for the GUT vscode extension.  If you think this might get merged at some  point I can  address the merge conflicts.   Also noticed I added package-lock.json to the gitignore which I now know was wrong to do.

PR #165, opened 2020-04-23
Add debugging support to the extension

### LINKED ISSUE #72: Add debuging support for GDScript
hello, I use godot but it is not useful for its built-in editor, I would like to know if you plan to make the groups appear, as well as if you plan to implement the debug. to track the correct functioning of the scripts

Razoric480:
# Godot Debugger 2.0 #

The previous codebase in pull request #146 was built while I didn't fully understand how the system was going to work at the end of the project. As a result, clumsy systems piled on top of clumsy systems and it proved to be buggy and hard to fix. After a period of dormancy where I left the project untouched, I decided to hop back on and fix.

The code has been refactored and cleaned up _significantly,_ and missing features have been added. As previous pull request, closes #72 .

## Supports ##

- Step-in/over/out
- Continue
- Pause (though since Godot pausing doesn't report a breakpoint, the editor doesn't change, but hitting pause again will resume)
- Locals, members and globals scopes for variables while broken
- Scope drilling and objects-within-objects, including for Object types.
- Exceptions breaking to not allow continue/step-in/step/resume
- Stack trace and jumping between those frames
- The ability to launch a game instance to debug with or to attach to a debugger after the fact, a game instance pointing at the currently opened TSCN, or to specify a TSCN file.
- Active scene tree is displayed and its nodes can be accessed in the Inspector
- Inspector allows to browse nodes and values, and edit primitives (int, float, strings, bools.)

README has been updated

## Unfixable ##

- Threads are not supported, though they don't currently work in the official debugger either.

## To be done ##

Set up a branch in preparation for [4.0](https://github.com/godotengine/godot/pull/36244) - At least, if the commands and their layouts have changed.


Geequlim:
Thanks!

PR #178, opened 2020-05-11
Fix scene_file and launch failure bug

Razoric480:
- scene_file used an unnecessary call to path.relative that prevented proper use.
- If the debugger failed to launch, it would fail on future attempts.

### DIFF
--- a/src/debugger/server_controller.ts
+++ b/src/debugger/server_controller.ts
@@ -95,6 +95,26 @@ export class ServerController {
 	) {
 		this.debug_data = debug_data;
 
+		if (launch_instance) {
+			let godot_path: string = utils.get_configuration("editor_path", "godot");
+			let executable_line = `"${godot_path}" --path "${project_path}" --remote-debug ${address}:${port}`;
+			if (launch_scene) {
+				let filename = "";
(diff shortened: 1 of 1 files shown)


src/debugger/server_controller.ts
```diff
@@ -95,6 +95,32 @@ export class ServerController {
...
+			let godot_path: string = utils.get_configuration("editor_path", "godot");
+			let executable_line = `"${godot_path}" --path "${project_path}" --remote-debug ${address}:${port}`;
+			if (launch_scene) {
+				let filename = "";
+				if (scene_file) {
+					filename = scene_file;
+				} else {
+					filename = window.activeTextEditor.document.fileName;
+				}
+				if (path.extname(filename).toLowerCase() === ".tscn") {
```
Geequlim, line 108: How about scenes saved with `tres` `scn` or other extensions ?

Razoric480: Those are a thing? =o I suppose I could just get rid of the check entirely, and if the file can't launch, then it will just fail to launch.

Geequlim:
Thanks!

PR #182, opened 2020-05-19
Add ability for user to set scene file on right click in explorer

ryerrappa:
Right clicking on a .tscn file will allow you to press "Set as Scene File" and set that file as the debugger's scene_file ([line 187 of package.json](https://github.com/godotengine/godot-vscode-plugin/blob/master/package.json#L187-L191))

This is done by setting an additional workspace configuration "godot_tools.scene_file_config". 

-If godot_tools.scene_file_config is set through right click, it overrides "scene_file" found in     launch.json. The value of "scene_file", however, remains untouched.

-Right Clicking the same .tscn file as scene_file_config and pressing "Set as Scene File" will unset scene_file_config

### DIFF
--- a/package.json
+++ b/package.json
@@ -287,7 +287,6 @@
 			"explorer/context": [
 				{
 					"command": "godot-tool.set_scene_file",
-					"when": "resourceLangId == properties && resourceExtname == .tscn",
 					"group": "2_workspace"
 				}
 			]
--- a/package.json
+++ b/package.json
@@ -74,6 +74,10 @@
 					"light": "resources/light/icon_edit.svg",
 					"dark": "resources/dark/icon_edit.svg"
 				}
+			},
+			{
+				"command": "godot-tool.set_scene_file",
+				"title": "Set as Scene File"
 			}
 		],
 		"configuration": {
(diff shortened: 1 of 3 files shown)


src/debugger/debug_session.ts
```diff
@@ -236,6 +240,7 @@ export class GodotDebugSession extends LoggingDebugSession {
 			args.launch_scene,
 			args.scene_file,
```
Razoric480, line 241: Could be more concise with
```javascript
get_configuration("scene_file_config", "") || args.scene_file
```
without the `if` and replacing the contents of args.scene_file.

package.json
```diff
@@ -269,6 +283,13 @@
 					"when": "view == inspect-node && viewItem == editable_value",
 					"group": "inline"
 				}
+			],
+			"explorer/context": [
+				{
+					"command": "godot-tool.set_scene_file",
+					"when": "resourceLangId == properties && resourceExtname == .tscn",
```
Razoric480, line 290: There was a similar comment here where forcing `.tscn` files to be the only thing you can run to be an issue when I first implemented launching a scene file, to be able to launch scenes that do are _not_ `.tscn` files.

Razoric480:
This seems to work pretty well, actually. More convenient than manually setting it in launch.json.

I added a couple of comments, though I'm not an official reviewer.

src/debugger/debug_session.ts
ryerrappa: That's pretty cool! Will do. Unfortunately have to sleep for now.

Calinou (approved):
Thanks!

camsjams:
This is a great addition, but it seems to have been placed at the top of the context menu list, which break a lot of workflows when removing and adding folders to workspaces.

Calinou:
@camsjams Feel free to open a pull request to move it down :slightly_smiling_face:

PR #190, opened 2020-06-04
Remove `extends` snippet which no longer works since Godot 3.0

Calinou:
- Remove `set_process(true)`/`set_process_input(true)` snippets since these are enabled automatically in Godot 3.0 onwards.


Geequlim:
Thanks.

PR #197, opened 2020-06-13
Set up continuous integration using GitHub Actions

Calinou:
The extension will be built and uploaded to artifacts so it can be tested easily by users.

This also fixes warnings reported by TSLint.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,26 @@
+name: Continuous integration
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Install Node.js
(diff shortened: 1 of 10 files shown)


Calinou:
@Geequlim Why not push it? It ensures everyone's using the same version of all the dependencies.

Geequlim:
I think the dependencies version are defined in package.json

Calinou:
> I think the dependencies version are defined in package.json

Yes, but only partially. package.json will only constrain the versions for direct dependencies, not indirect dependencies. Also, [it allows installing newer semver-compatible versions](https://github.com/godotengine/godot-vscode-plugin/blob/master/package.json#L275-L294) due to the use of the caret symbol (`^`) in version ranges.

Razoric480:
Having just run into an issue where the code is different between 2 of my machines because my `package-lock.json` was different, I can assure that having a package-lock.json pushed to the repo to be a good thing.

PR #218, opened 2020-07-18
Add check for PowerShell 7 executable name

### LINKED ISSUE #201: Godot start command does not work with Powershell 7.x
The GodotTools command for starting the Godot editor does not work with Powershell 7.x:
Screenshot of invalid command

To make the command valid, `& ` must be prepended.
Screenshot of corrected command

a-type:
Fixes #201

### DIFF
--- a/src/godot-tools.ts
+++ b/src/godot-tools.ts
@@ -70,10 +70,10 @@ export class GodotTools {
 					let cmdEsc = `"${cmd}"`;
 					if (process.platform === "win32") {
 						const POWERSHELL = "powershell.exe";
-						const POWERSHELL_7 = "pwsh.exe";
+						const POWERSHELL_CORE = "pwsh.exe";
 						const shell_plugin = vscode.workspace.getConfiguration("terminal.integrated.shell");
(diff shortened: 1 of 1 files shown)


taylorjoshuaw:
I'd probably name that constant POWERSHELL_CORE rather than POWERSHELL_7

taylorjoshuaw:
I just thought of something -- Powershell Core isn't tied to Windows like the old Powershell was. It might be a good idea to move your change out of the win32 platform check and not expect a ".exe" extension at the end . Sorry I didn't catch that before.

a-type:
@taylorjoshuaw good point! I can do that.

Calinou:
Is this mergeable as-is? Maybe we can continue improving this later.

taylorjoshuaw:
@Calinou Good to see you around here (I was about to ask if you needed some help merging PR's). It should be good to merge. I think the amount of people using this plugin and Powershell Core as their integrated shell on non-Windows platforms is probably pretty limited.

Calinou (approved):
Thanks!

PR #229, opened 2020-09-14
Fix no header, non-TCP compliant msg for natives

### LINKED ISSUE #210: Selecting an item in the List native classes of Godot list fails to get docs and causes future hovers to get stuck
1. Open a godot project, connect to LSP
2. Hover over a `print()` statement, see that it brings up the documentation for it normally.
3. Open `List native classes of Godot` command and press enter on one of them. The documentation page doesn't come up.
4. Hover over the same `print()` statement, and it's stuck at `Loading...` until reconnected.

The extension outputs the following when debugging it:
```
[godot-tools][2020-06-27T17:39:41.806Z] [server] {"error":{"code":-32700,"message":"Parse error"},"id":null,"jsonrpc":"2.0"} 
loggger.js:36
Received response message without id: Error is: 
{
    "code": -32700,
    "message": "Parse error"
}
```

From then on, it receives the textDocument/hover command, but doesn't do anything with it until it's cancelled by the $/cancelRequest command.

This is latest master on 3.2.2 with tcp/ip.
This issue does not happen on 3.2.1 with websocket.

Razoric480:
The way messages are sent to the LSP server via TCP is slightly different between 3.1 websocket and 3.2 TCP/IP, requiring a header indicating the length of the message and a JSON object with a signature that requires a params: key that was lacking, and no ID.

Fixes #210 

Also adds the Find widget to the native symbol pages that come up so you can CTRL+F through a page.

### DIFF
--- a/src/lsp/NativeDocumentManager.ts
+++ b/src/lsp/NativeDocumentManager.ts
@@ -1,81 +1,113 @@
-import * as vscode from 'vscode';
+import * as vscode from "vscode";
 import { EventEmitter } from "events";
 import { MessageIO } from "./MessageIO";
 import { NotificationMessage } from "vscode-jsonrpc";
 import * as Prism from "../deps/prism/prism";
 import * as marked from "marked";
-import { Methods, NativeSymbolInspectParams, GodotNativeSymbol, GodotNativeClassInfo, GodotCapabilities } from './gdscript.capabilities';
+import { get_configuration } from "../utils";
+import {
+	Methods,
+	NativeSymbolInspectParams,
+	GodotNativeSymbol,
+	GodotNativeClassInfo,
+	GodotCapabilities,
+} from "./gdscript.capabilities";
 marked.setOptions({
 	highlight: function (code, lang) {
 		return Prism.highlight(code, GDScriptGrammar, lang);
-	}
+	},
 });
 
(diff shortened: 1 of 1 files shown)


Shatur95:
Can confirm, works for me.

Shatur95:
Is it possible to use text search in such documentation? In the previous versions text in documentation was not searchable too.

Razoric480:
You can search by class name, but not by the contents. We don't have access to the contents of the docs to search _with_.

Godot sends a message containing its capabilities, which is a list of native classes and symbols, and the VSCode client stores them. When you ping the list, VSCode sends a message back to the language server asking it about a specific native symbol. Godot responds with the docs page.

To make it searchable via text, we would need to either include a copy of the docs in the extension, or bombard the local server with a request for every native class and store them.

Shatur95:
> EDIT: Unless you mean just being able to CTRL+F through one of the document you receive, in which case, maybe it's possible, I'll investigate.

Yes, this is just what I mean.

But this PR can be merged I think.

Razoric480:
Actually I just figured out it's very to do. I'll add it to the PR while it's open.

Shatur95:
Thank you a lot, much more convenient now!

Geequlim:
Thanks

PR #236, opened 2020-10-11
Check for updates to GitHub Actions every weekday

jlosito:
Now that dependabot is native to GitHub and this project uses actions, it would be nice to always keep them up to date. This PR will allow dependabot to check for updates to any actions uses and submit PRs with version bumps automatically.

https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/keeping-your-actions-up-to-date-with-github-dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"


jlosito:
@NathanLovato this is the same as the PR linked below. I figured it would be easier to maintain actions which are all on the same version across each of the projects.

https://github.com/godotengine/godot-docs/pull/4236

Calinou (approved):
Thanks!

Issue #231, opened 2020-09-20
Opening scripts from Godot doesn't open them in the VS Code workspace window when using a VS Code workspace


mickdekkers:
I created a VS Code workspace for my project so I can store [project-specific VS Code settings](https://code.visualstudio.com/docs/getstarted/settings) and [extension recommendations](https://code.visualstudio.com/docs/editor/multi-root-workspaces#_extension-recommendations) alongside the project files and enable/disable certain extensions for the workspace.

### Steps to reproduce

1. Create a new project in Godot
2. Open the project folder in VS Code (either manually or through Godot)
3. File -> Save Workspace As... -> (`<anything>.code-workspace`) -> Save

The project folder is now open as part of the workspace we just saved.

4. Create a new node and attach a script to it in Godot
5. Observe as a new VS Code window is opened for the project folder instead of the existing window being used

The same issue occurs when opening an existing script from Godot. Also, if you close the new VS Code window and try to open the script from Godot again, it will reopen the window. Opening scripts from Godot never opens them in the workspace window, even if the script is already open in that window.

Editing scripts from the workspace window works fine, they are just not opened there from Godot.

Restarting VS Code or Godot does not affect the issue.

#### Versions

Windows 10
Godot version 3.2.3.stable.official
VS Code version 1.49.1
godot-tools version 1.1.2

#### Settings

`"godot_tools.editor_path": "C:/software/Godot_v3.2.3-stable_win64.exe"`

I set Godot's external editor `Exec Path` and `Exec Flags` settings according to the README.

linkpy:
Removing `{project}` from the `Exec Flags` option fixes the issue for me. I dont think the issue is from the extension or Godot itself, I think it's related to how VS Code handles multiple instances.

What you can do is modify the `Exec Flags` option like so `{project}/<anything>.code-workspace --goto {file}:{line}:{col}`. For me it  works as intended and only use the window which has that workspace open.

SbhnP:
Is there any update on this?

I need files to be opened in the active vscode workspace window,Which is a repo which contains Godot project folder in it's sub folders. but Godot opens the file a new window . 

`text_editor/external/use_external_editor` is `--goto {file}:{line}:{col}`

Edit: I just found out disabling `dotnet/editor/external_editor` fixes the problem.

PR #241, opened 2020-10-22
[GH-239] Add godot_tools.project_path property

### LINKED ISSUE #239: Add extension setting to specify project file path so that the vs workspace can be treated as a godot workspace
I'm in the progress of building out go binding for godot here: [godot-go](https://github.com/godot-go/godot-go). I was wondering if you guys would be interested in adding support for specifying a project file not at the root of the workspace?

basically something like this:
```
godot --path test/project/
```

without this feature, the extension doesn't treat my vscode workspace as a godot workspace and i get the ["Current workspace is not a Godot project"](https://github.com/godotengine/godot-vscode-plugin/blob/60cb0c57fcbafe37f6389d9b3df157c95943e8f3/src/godot-tools.ts#L53) error.

pcting:
*Bugsquad edit: This closes #239.*

### DIFF
--- a/src/godot-tools.ts
+++ b/src/godot-tools.ts
@@ -11,7 +11,6 @@ export class GodotTools {
 
 	private context: vscode.ExtensionContext;
 	private client: GDScriptLanguageClient = null;
-	private workspace_dir = vscode.workspace.rootPath;
 	private project_file = "project.godot";
 	private connection_status: vscode.StatusBarItem = null;
 
@@ -49,12 +48,14 @@ export class GodotTools {
 
 		return new Promise((resolve, reject) => {
 			let valid = false;
-			if (this.workspace_dir) {
-				let cfg = path.join(this.workspace_dir, this.project_file);
+			const projectPath = path.join(vscode.workspace.rootPath, get_configuration("project_path", ""));
+			if (projectPath) {
+				let cfg = path.join(projectPath, this.project_file);
 				valid = (fs.existsSync(cfg) && fs.statSync(cfg).isFile());
 			}
 			if (valid) {
-				this.run_editor(`--path "${this.workspace_dir}" ${params}`).then(()=>resolve()).catch(err=>{
+				const args = `--path "${projectPath}" ${params}`;
+				this.run_editor(args).then(()=>resolve()).catch(err=>{
 					reject(err);
 				});
 			} else {
@@ -72,7 +73,7 @@ export class GodotTools {
 		else {
 			scene_config = right_clicked_scene_path
 		}
-		
+
 		set_configuration("scene_file_config", scene_config);
 	}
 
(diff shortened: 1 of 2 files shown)


Calinou:
Out of curosity, what is this used for?

src/godot-tools.ts
```diff
@@ -11,7 +11,6 @@ export class GodotTools {
 
 	private context: vscode.ExtensionContext;
 	private client: GDScriptLanguageClient = null;
-	private workspace_dir = vscode.workspace.rootPath;
```
pcting, line 14: I know nothing about vscode extensions. I originally wanted to call `get_configuration` here, but I decoded to move the `get_configuration` into each method call below because I wanted to make sure we had the latest value in `settings.json`.

pcting: Ah, i found it here: https://code.visualstudio.com/api/references/vscode-api#workspace
```
onDidChangeConfiguration: Event<ConfigurationChangeEvent>
```

pcting:
just checking in to get a status on this PR

Calinou:
@linkpy Could you review this PR, please?

linkpy:
Is a new option really necessary for this ? 

It seems like just switching from `vscode.workspace.rootPath` to `vscode.workspace.workspaceFolders` would be enough in order to find the correct folder containing the `project.godot` file.

Also, the issue behind this PR is pretty close to #231.

pcting:
my use case is to be able to continue developing my [godot-go](https://github.com/godot-go/godot-go) bindings library. without this feature, the `godot.project` file would have to be moved to the root of the repository to have a seamless experience with the plugin. i think the alternative would be:
1. open a separate window for the `godot-go/test/` path so that the `workspaceFolder` contains the `godot.project` at the root. The downside is you'll be switching between 2 separate vscode windows to do development plus back to the godot editor; this is not ideal.
2. start a vscode workspace such 2 folders are mounted in that workspace: 1) the godot-go/test/ as the first folder and 2) the godot-go/ as the second folder. the first folder must contains the godot project file; otherwise, you'll get an error if you try to open up the godot editor with the plugin otherwise. this is an acceptable alternative as most developers will be working with on a godot project with 1 folder. aside from having the `godot/test/` folder visible from 2 places, it will work.

side note: rootPath has be [deprecated](https://github.com/microsoft/vscode/wiki/Adopting-Multi-Root-Workspace-APIs#eliminating-rootpath)

a sample vscode workspace file:
```
{
	"folders": [
		{
			"path": "."
		},
		{
			"path": ".."
		}
	]
}
```

linkpy:
Ok from what I understood initially it was an issue with a workspace, not with a single folder.
Well in that case, you should push a new commit which switches `rootPath` to `workspaceFolders` and then we should be able merge it.

DaelonSuzuka:
I think this issue has generally been solved.

PR #247, opened 2020-10-28
Fixed the syntax of the escaped characters in strings

### LINKED ISSUE #214: Escaped quote causes syntax coloring to stop?
Sprite-0001

	var x = line.split("\"")

linkpy:
Fixes #214


Calinou (approved):
Thanks!

PR #262, opened 2021-03-09
Fix typo in snippets: "decleration" -> "declaration"

LightningAA:

PR #263, opened 2021-03-09
Fix typo "guii" -> "gui" and "decleration" -> "declaration"

LightningAA:


Calinou:
I think the `guii` *might* be intentional, since you don't want it to appear every time you enter "gui". I'm not 100% sure though.

Razoric480:
I think Calinou's got it, yeah. Looks like it was intentional to me to allow the use of a variable called gui (which would be fairly common) without triggering VSCode's snippets.

The other looks like a typo to me.

Calinou:
> The other looks like a typo to me.

The "declaration" typo has been fixed in https://github.com/godotengine/godot-vscode-plugin/pull/262 already :slightly_smiling_face:

LightningAA:
Alright, sound good

Issue #275, opened 2021-06-13
[Discussion] GDScript grammar file


Vlad-Zumer:
Looking through the issues, seems there are quite a few problems related to the grammar we use to highlight GDScript code.
Unfortunately I have been unable to find in the documentation the grammar for GDScript. 
Should someone from the core project provide and keep up to date a file containing the grammar, the syntax highlight could be  updated and made less prone to errors.

Issues regarding highlighting:

- #139
- #133 
- #256 
- #158 (possibly)
- #219 (possibly)

Calinou:
There's already a GDScript grammar file: https://docs.godotengine.org/en/latest/development/file_formats/gdscript_grammar.html

Vlad-Zumer:
The problem with both both those is that they aren't exhaustive and they are missing references to some language specifications.
Even though some of these can be implied, it would be better if we had a concrete spec.

E.g.

- Strings, multi line strings, formatted strings, escaped strings
- builtin functions
- `$` shorthand

Another solution I am looking into is having the language server generate the highlighting as well. This will however requires it to be a stand-alone application we can start via VSCode and have Godot link to it in the same way. With the current architecture `.gd` files would only have syntax highlighting if Godot is running

New architecture example
```
VSCode <---LSP---> GDScript Language Server <------> Godot
```

Calinou:
> The problem with both both those is that they aren't exhaustive and they are missing references to some language specifications.

GDScript doesn't have a formal language specification, but I can tell you some caveats about the things you mentioned.

> - Strings, multi line strings, formatted strings, escaped strings

Strings can use the following forms:

```gdscript
"One"
'Two'

"""
Three
"""

'''
Four
'''

"""Multi-line strings do not *have* to be on multiple lines, and can contain "quoted" strings without requiring escapes (unless there are 3 quotes or more)."""
```

> - builtin functions

A list of built-in functions can be found here: https://github.com/godotengine/godot/blob/f9521d35cb4fc44d6a54e50bc7554e523c2b93ef/modules/gdscript/gdscript_functions.cpp#L46-L137

> - $ shorthand

The following paths are valid:

```
# Node names must be valid identifiers if they are not quoted. This means they can't start with a number:
$SomeNode
$"2SomeNode"

# No quotes required for paths with only node names and slashes, but they are allowed.
$SomeNode/SomeChildNode

# Quotes are required for paths that contain `..` (double or single quotes).
$"../Example"

[...]

PR #276, opened 2021-06-15
Change `$` node path shorthand regex to match only valid paths

Vlad-Zumer:
- Changed regex to allow only valid paths as discussed in #275 
- Added a `README.md` with useful information for debugging and creating regexes for the GDScript syntax.

### DIFF
--- a/syntaxes/README.md
+++ b/syntaxes/README.md
@@ -8,4 +8,4 @@ When creating regexes it is sometimes useful to test against a data set.
 
 [Regex101](https://regex101.com/) allows for pretty good match testing, unfortunately it does not support the regex (ruby) version that TextMate uses.
 
-[ExtendsClass's](https://extendsclass.com/regex-tester.html#ruby) regex engine does support the ruby the variant but the match testing experience is not as good. However it does automatically generate an FSM for your regex, which is very useful in debugging.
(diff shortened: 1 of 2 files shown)


syntaxes/README.md
```diff
@@ -0,0 +1,11 @@
+# Helpful resources
+
+You can find the regex syntax that VSCode uses [here](https://macromates.com/manual/en/regular_expressions).
+
+## Creating and debugging regex
+
+When creating regexes it is sometimes useful to test against a data set.
+
+[Regex101](https://regex101.com/) allows for pretty good match testing, unfortunately it does not support the regex (ruby) version that TextMate uses.
```
Calinou, line 9:
```suggestion
[Regex101](https://regex101.com/) allows for pretty good match testing. Unfortunately, it does not support the regex (Ruby) version that TextMate uses.
```

syntaxes/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+[ExtendsClass's](https://extendsclass.com/regex-tester.html#ruby) regex engine does support the ruby the variant but the match testing experience is not as good. However it does automatically generate an FSM for your regex, which is very useful in debugging.
```
Calinou, line 11:
```suggestion
[ExtendsClass's](https://extendsclass.com/regex-tester.html#ruby) regex engine does support the Ruby variant, but the match testing experience is not as good. However, it automatically generates an FSM for your regex, which is very useful when debugging.
```

Calinou (changes requested):

Calinou (approved):
Thanks!

PR #278, opened 2021-06-17
Fix conditional breakpoints being parsed as regular breakpoints

### LINKED ISSUE #267: Visual Studio incorrectly pauses on conditional breakpoints.
I'm using visual studio code with godot-tools extension. To test this out, I set a conditional breakpoint with a condition that's impossible to achieve.

The game incorrectly freezes on the breakpoint, and from what I seen in videos, there's supposed to be a yellow marker outline on the breakpoint, but that's not the case. My only options when it freezes is pause, restart, or start which is a sign that it hasn't actually hit the breakpoint.

When I disable the breakpoint the game runs fine, so it's not the coding.

EDIT: I changed the title, because the freezing isn't happening anymore, but the incorrect pausing is.

Razoric480:
Despite `supportsConditionalBreakpoints` being set to `false`, seems the debug adapter still expects the debugger to explicitly filter out conditional breakpoints, or continue treating them as regular breakpoints (which makes the presence of `supportsConditionalBReakpoint` kind of pointless? I dunno.)

Until Godot supports conditional breakpoints, I'm opting to just remove them entirely as there's no feasible way of hooking into the debugger so there's no confusion. Now they'll be ignored.

Fixes #267

### DIFF
--- a/src/debugger/debug_session.ts
+++ b/src/debugger/debug_session.ts
@@ -298,7 +298,10 @@ export class GodotDebugSession extends LoggingDebugSession {
 			});
 			client_lines.forEach((l) => {
 				if (bp_lines.indexOf(l) === -1) {
-					this.debug_data.set_breakpoint(path, l);
+					let bp = args.breakpoints.find((bp_at_line) => (bp_at_line.line == l));
+					if(!bp.condition) {
+						this.debug_data.set_breakpoint(path, l);
+					}
 				}
 			});
 
--- a/src/debugger/debug_session.ts
+++ b/src/debugger/debug_session.ts
@@ -299,7 +299,7 @@ export class GodotDebugSession extends LoggingDebugSession {
 			client_lines.forEach((l) => {
 				if (bp_lines.indexOf(l) === -1) {
 					let bp = args.breakpoints.find((bp_at_line) => (bp_at_line.line == l));
-					if(!bp.condition) {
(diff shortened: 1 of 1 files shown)


src/debugger/debug_session.ts
```diff
@@ -298,7 +298,10 @@ export class GodotDebugSession extends LoggingDebugSession {
 			});
 			client_lines.forEach((l) => {
 				if (bp_lines.indexOf(l) === -1) {
-					this.debug_data.set_breakpoint(path, l);
+					let bp = args.breakpoints.find((bp_at_line) => (bp_at_line.line == l));
+					if(!bp.condition) {
+						this.debug_data.set_breakpoint(path, l);
+					}
```
Calinou, line 304:
```suggestion
					if (!bp.condition) {
						this.debug_data.set_breakpoint(path, l);
					}
```

AlexFolland:
The issue this fixes affects my workflow.  Thank you for your work on this.  Is there anything preventing this from being merged?

Calinou (approved):
Thanks!

AlexFolland:
That would be lovely.  Thank you for merging this!

Geequlim:
I remember you have permission to publish.

Hugo Locurcio ***@***.***>于2021年11月3日 周三22:16写道：

PR #279, opened 2021-06-19
Fix "static func" indent error

OptimisticQuan:
When declare `static func` and press enter key, two indents will be generated. 

Fix by add "static func" pattern

### DIFF
--- a/configurations/gdscript-configuration.json
+++ b/configurations/gdscript-configuration.json
@@ -23,7 +23,7 @@
         ["{", "}"]
     ],
     "indentationRules": {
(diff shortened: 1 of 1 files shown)


Calinou:
@NathanLovato I remember hearing that this wouldn't be the proper fix (and that this should be fixed in the Godot editor LSP instead). Do I remember this correctly?

NathanLovato:
@Calinou Indentation and coloring are independent of the language server - it's editor plugin stuff - so this seems the right fix at a glance. Just needs someone to test it.

akien-mga (approved):

PR #313, opened 2022-01-26
Update dependencies, recommend local TypeScript and vsce

Calinou:
Local package installations should be favored over global installations to avoid conflicts when working on multiple projects on the same system.

Only dependencies that did not cause the build to break were updated.

- Add `package` npm script to run local vsce installation.

I built and installed the extension locally and can confirm it still works.


akien-mga (approved):

akien-mga:
Thanks!

Issue #296, opened 2021-09-01
"Continue" doesn't work for multiple breakpoints in the same script


olestourko:
OS: Ubuntu 20.04
Godot: 3.3.3
VSCode: 1.59.1

When setting two breakpoints within the same script, only the first one gets hit. When I press "continue" the next one doesn't get triggered. Continue **does** work correctly for breakpoints in separate scripts.

![Screenshot_2021-09-01_18-14-37](https://user-images.githubusercontent.com/3958783/131752518-02b447d4-6df8-4480-8b10-6868d090aa81.png)

^ Only the one on L12 works. L14 not triggered on Continue. L14 does work if I remove the L12 breakpoint.

olestourko:
My `launch.json` file:

```
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "name": "GDScript Godot",
            "type": "godot",
            "request": "launch",
            "project": "${workspaceFolder}",
            "port": 6007,
            "address": "127.0.0.1",
            "launch_game_instance": true,
            "launch_scene": false
        }
    ]
}
```

gabstv:
It's happening to me too.  The debugger just gives up mid file.

raulsntos:
Closed by #324

PR #324, opened 2022-02-24
Fix "Continue" for multiple breakpoints in the same script

kurtlachmann:
Fixes issue #296

When starting a new debug session a `setBreakPointsRequest()` is triggered for each file that contains breakpoints.
However at that time `this.project_path` is still undefined. It will only be initialized during the `launchRequest()`.
Trying to pass `undefined` into `path.relative()` will cause a type error which means the `setBreakPointsRequest()` will fail and exit after the first breakpoint of each file.

With this fix we skip the respective lines during initialization.
At that time the `project_path` is still unknown so we can't construct the relative path.
Also the Mediator isn't connected yet, so these lines wouldn't do anything anyway.

### DIFF
--- a/src/debugger/debug_runtime.ts
+++ b/src/debugger/debug_runtime.ts
@@ -82,8 +82,9 @@ export class GodotDebugData {
 
 		bps.push(bp);
 
-		let out_file = `res://${path.relative(this.project_path, bp.file)}`;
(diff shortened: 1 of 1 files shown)


Calinou (approved):
Thanks!

PR #325, opened 2022-02-24
Use the LSP defined SymbolKind enum and fix marked

### LINKED ISSUE #321: Missing properties, methods and signals  in "List Native Classes in Godot"
Using  ```Godot_v3.4.3-rc2_win64``` 
and ```vscode 1.64.2 (system setup) OS: Windows_NT x64 6.1.7601 ```  +  ```godot-tools v1.1.3```

Tested also with ```Godot_v3.4.2-stable_win64``` - the same result.

raulsntos:
Since Godot uses the `SymbolKind` enum defined by the [LSP protocol](https://microsoft.github.io/language-server-protocol/specification#textDocument_documentSymbol) instead of the one defined by [VSCode](https://github.com/microsoft/vscode/blob/459ffea03c117042c17c6a3c02f066c4af5db555/src/vs/editor/common/languages.ts#L1175-L1202).
This enum was changed in Godot in https://github.com/godotengine/godot/pull/50913 and was backported to 3.x in https://github.com/godotengine/godot/pull/50914.

Also fixes `marked` being called incorrectly to parse codeblocks.

Fixes #321

### DIFF
--- a/src/lsp/NativeDocumentManager.ts
+++ b/src/lsp/NativeDocumentManager.ts
@@ -1,4 +1,5 @@
 import * as vscode from "vscode";
+import * as ls from "vscode-languageclient";
 import { EventEmitter } from "events";
 import { MessageIO } from "./MessageIO";
 import { NotificationMessage } from "vscode-jsonrpc";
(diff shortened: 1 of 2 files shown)


Calinou (approved):
Thanks!

PR #327, opened 2022-02-25
Fix restart of debugging sessions

### LINKED ISSUE #282: Using the restart debugging button(next to stop button) restarts the game, but stops debugging.

kurtlachmann:
We have to send a response to the `configurationDoneRequest` or else vscode will run into a timeout and disconnect the debug adapter. For some reason this timeout does not apply when launching the session initially. That's why stopping and then starting a new session worked before.

Fixes #282


Calinou (approved):
Thanks!

Issue #196, opened 2020-06-13
Add syntax highlighting for `.tscn/.tres/.godot` files


Calinou:
This is low-priority, but it'd be nice to have a dedicated highlighter for Godot scenes and resources. The automatically-detected "Properties" syntax highlighting does its job decently, but a specially tailored highlighter could do better. For instance, it could handle slashes in property names and highlight built-in Godot types.

See also #70 which is about `.shader` files.

Calinou:
> The shader language is probably similar enough to use an existing grammar with minimal problems, but autocompletion or inline documention will need a language server or something.

There is no LSP for the Godot shader language yet, although the list of built-in functions is known in advance (and is similar to GLSL). Using an existing GLSL grammar as a base should work well :slightly_smiling_face: 

For `.tscn`, I would recommend using a TOML or INI grammar as a base.

DaelonSuzuka:
I put together a grammar that works on `.tscn`, `.tres`, `.import`, `.gdnlib`, `.gdns`, and `.godot` files.

The colors are mostly similar to the gdscript grammar, and if anybody thinks it would look better tagged a different way I can certainly swap them around.

<details>
<summary>GDResource Grammar</summary>


![Code_7lSgUNU4wS](https://user-images.githubusercontent.com/18042232/156536432-be825960-f103-4380-b825-ab4345586a93.png)


</details>

Godot 4 resource files appears to work, except for missing the new type names(Vector2i, PackedStringArray, PackedInt32Array, etc). Is there a master list of the new types somewhere? If somebody could point me to it, I can add them easily.

<details>
<summary>GDResource for Godot 4</summary>


![Code_SxUM3xBNFt](https://user-images.githubusercontent.com/18042232/156537477-f4c7a4c7-9872-4cdf-8679-3a4b517fccf8.png)


</details>

And finally, here's extremely simple GDShader support. I basically just set up gdshader as an alias to glsl.

I also configured highlighting for gdshader code embedded in a gdresource file. (The right side of the below screenshot.)

<details>
<summary>GDShader Grammar</summary>

![Code_QDBE2dnwgT](https://user-images.githubusercontent.com/18042232/156553682-79f2d375-aee6-4022-833e-1fe594150ac7.png)


</details>

Calinou:
@DaelonSuzuka Looks great! Feel free to open a pull request for this :slightly_smiling_face:

DaelonSuzuka:
[...]

Issue #334, opened 2022-03-30
Syntax highlighting partially fails for signal forward declarations with parameters 


knochenhans:
### Godot version

3.4.4

### VS Code version

1.65.2

### Godot Tools VS Code extension version

1.1.3

### System information

Arch Linux

### Issue description

Syntax highlighting fails for me when I add parameters to signal forward declarations, the colour is set to standard grey for the following text up to the first function definition. I attached a GIF screen recording that shows the difference when I add/delete the parameter.

![Peek 2022-03-30 23-41](https://user-images.githubusercontent.com/5293125/160935880-8bd35918-44f3-48c6-be27-b2e5bcacf987.gif)

### Steps to reproduce

Add a signal forward declaration with a parameter.

DaelonSuzuka:
I'm currently overhauling the syntax highlighting and I'll make sure this is addressed.

knochenhans:
Many thanks :) by the way, this also seems to happen for return type annotations:
![grafik](https://user-images.githubusercontent.com/5293125/162722979-f28a66ea-f395-438e-aa69-6131b4b9ac3d.png)

DaelonSuzuka:
No problem. There legitimately was a known issue with signal declarations, so it wasn't a stretch to think there could be more problems.

PR #342, opened 2022-04-06
Overhaul syntax highlighting

DaelonSuzuka:
This satifies #196 by adding grammars to support `.tscn`, `.tres`, `.import`, `.gdnlib`, `.gdns`, and `.godot` files.

The GDResource grammar has special string highlighting in a number of cases:

- `TypeName`s like a Node or resource's `type` field are green like regular use of a type 
- NodePath like a Node's `name` field or the argument of a NodePath constructor are orange
- resource paths are yellow
- embedded GDScript resources (check built-in when adding a script) have full syntax highlighting
- embedded shader resources have full syntax highlighting

<details>
<summary>GDResource screenshots </summary>

 `.tscn`
Code_q5uqtRrnLc

`.import`
Code_D4lIF9zfMt

`.gdns`
Code_M7h4gMwebU

</details>

There are many improvements to the existing GDScript grammar, including:

- adding Godot 4 classes and keywords
- separate tagging of control flow keywords(if, else, for, etc)
- correct handling of many previously missed edge cases
- fixed regressions caused by recent bug fixes

I made a potentially controversial decision to give special highlighting to string literals that are used as `NodePath`s. This include the parameters of `get_node()` and `has_node()`, as well as the path specified using the $ operator. 

There is very basic support for `.shader` files. Essentially, they're just highlighed using glsl rules, but it's better than nothing, and there's now a hook to come back and develop a better custom grammar.

The gdresource grammer for `.tscn` and `.tres` files also correctly highlights embedded shader code and embedded gdscript.

I added a folder for example files that can be used to quickly check that special cases are handled correctly. Currently the only file is `syntaxes\examples\gdscript1.gd`, but I want to add similar examples for the other files types(including a `gdscript2.gd` example). I would greatly appreciate contributions to these test files, because a better suite of test files will greatly reduce the risk of syntax highlighting regressions.

~~This PR is marked as a draft because there were upstream changes to the gdscript syntax that I need to take the time to merge by hand, since the nested regex rules can interact in very surprising ways.~~


atirut-w:
> * adding Godot 4 classes and keywords

Wouldn't that conflict with Godot 3's grammar?

Calinou:
> Wouldn't that conflict with Godot 3's grammar?

I think the best decision for now would be to have support for *both* Godot 3 and Godot 4 keywords, preferring Godot 3 in case of a conflict. This ensures the extension can support both Godot 3.x and Godot 4.0 projects for a while.

DaelonSuzuka:
> > * adding Godot 4 classes and keywords
> 
> Wouldn't that conflict with Godot 3's grammar?

That does not appear to be case. Some expiriments with matching the new @decorator syntax suggest that it will also be compatible with both. 

However, walking this line is one of the reasons I want to develop a suite of test files, so we can all be more confident that regressions don't make it out to users. 

> I think the best decision for now would be to have support for _both_ Godot 3 and Godot 4 keywords, preferring Godot 3 in case of a conflict. This ensures the extension can support both Godot 3.x and Godot 4.0 projects for a while.

I think it's very likely that both languages can be fully supported by one grammar. 

IF there ends up being a serious incompatibility, the most robust solution might be to support as big a common subset as possible with the grammar, and add tagging support to the language server, since the editor will know for sure which version is being parsed. (I will eventually work on this anyways, because I want to have semantic highlighting based on context.)

DaelonSuzuka:
I'm marking this ready for review even though I could easily spend another week polishing.

Attached are some screenshots of the test file I've been using to develop the gdscript grammar. On the left is 1.2.0, on the right is this PR.

<details>
<summary>Test File Screenshots</summary>

Lines 12-20: I've redone signal declarations, including a fix for #334 
Lines 23-55: The function declaration rule was fundamentally broken and had to be rebuilt

mGDy4MDlA4

Lines 68-71: Variable name rules were incorrectly tagging some of these as PascalClass names.
Lines 75-96: Special highlighting of everything that counts as a NodePath, including strings used as the parameter for get_node(), as_node() and NodePath()

kRNxrrnxSK

Lines 121-128: Type hint syntax was totally broken and overriding definition of dictionary literals. Call to int() was not properly detected as a type constructor. 

lGMJsV87ij

</details>

DaelonSuzuka:
Fixed the two issues from the last comment. 

I'm ready to not look at regex for a couple weeks, so I'm done polishing this for now unless somebody finds a mistake.

I'll make an issue soon to collect thoughts for future highlighting improvements.


<details>
<summary>Screenshots</summary>

Code_sJLy1dGGbO

Code_17r7YXU0Vp



</details>

syntaxes/GDScript.tmLanguage.json
```diff
@@ -1,351 +1,567 @@
...
-    },
-    "pascal_case_class": {
-      "captures": {
-        "1": { "name": "entity.name.type.class.gdscript" }
-      },
-      "match": "\\b([A-Z][a-zA-Z_0-9]*)\\b"
     }
-  }
-}
+}
```
Calinou, line 567:
```suggestion
}

```

syntaxes/GDShader.tmLanguage.json
```diff
@@ -0,0 +1,19 @@
...
+            "include": "source.glsl"
+        }
+    ],
+    "repository": {
+        "shader_type": {
+            "match": "(shader_type)",
+            "name": "storage.type.glsl"
+        }
+    }
+}
```
Calinou, line 19:
```suggestion
}

```

syntaxes/GDResource.tmLanguage.json
```diff
@@ -0,0 +1,329 @@
...
+                        },
+                        {
+                            "include": "#data"
+                        }
+                    ]
+                }
+            ]
+        }
+    }
+}
```
Calinou, line 329:
```suggestion
}

```

syntaxes/examples/gdscript1.gd
```diff
@@ -0,0 +1,145 @@
...
+			pass
+		elif false:
+			while true:
+				pass
+		else:
+			pass
+		
+		pass
+	
+
```
Calinou, line 145:
```suggestion

```

Calinou (approved):
Thanks!

Issue #337, opened 2022-03-31
Overaggressive formatting when adding new lines after an if statement


ZachIsAGardner:
### Godot version

3.4.4

### VS Code version

1.66.0

### Godot Tools VS Code extension version

1.1.3

### System information

Windows 10

### Issue description

![gdscript_extension_bug_2](https://user-images.githubusercontent.com/30502195/161106385-799a9321-e671-4225-899f-a919f0244891.gif)

As seen in the gif above, if I try to move a line down with either an "enter" key press or with the "move line down" shortcut command it indents it and adds it to the previous if statements block.

![gdscript_extension_bug_3](https://user-images.githubusercontent.com/30502195/161107146-07d69410-c0a7-4114-a00c-a11f2dbf86c4.gif)

Here's another example where I'm trying to copy paste something. It seems to work if I undo after pasting, but it's still a little annoying.

### Steps to reproduce

Try moving a line down after an if statement, or paste something after an if statement

was-games:
confirmed on the master branch

ZachIsAGardner:
I just tried this again it looks like it's fixed for Enter, but it's still behaving the same for **Move Line Down**, **Move Line Up**, and Pasting.

I'm using v1.3.1 of this plugin.

![5060](https://user-images.githubusercontent.com/30502195/172771271-9f582456-7320-49d7-b9a3-c050104cfdb6.gif)

PR #344, opened 2022-04-07
added strict onEnterRules for controlling indentation

was-games:
https://github.com/godotengine/godot-vscode-plugin/issues/337

### DIFF
--- a/configurations/gdscript-configuration.json
+++ b/configurations/gdscript-configuration.json
@@ -1,36 +1,93 @@
 {
     "comments": {
         "lineComment": "#",
-        "blockComment": ["\"\"\"", "\"\"\""]
+        "blockComment": [
+            "\"\"\"",
+            "\"\"\""
+        ]
     },
     "brackets": [
-        ["(", ")"],
-        ["[", "]"],
-        ["{", "}"]
+        [
+            "(",
+            ")"
+        ],
+        [
+            "[",
+            "]"
+        ],
+        [
+            "{",
+            "}"
+        ]
     ],
     "autoClosingPairs": [
-        ["'", "'"],
-        ["\"", "\""],
-        ["(", ")"],
-        ["[", "]"],
-        ["{", "}"]
+        [
+            "'",
+            "'"
+        ],
+        [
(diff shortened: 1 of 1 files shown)


configurations/gdscript-configuration.json
```diff
@@ -1,36 +1,93 @@
...
+            }
+        },
+        {
+            "beforeText": "^\\s*$",
+            "action": {
+                "indent": "none"
+            }
+        }
+    ]
+}
```
Calinou, line 93:
```suggestion
}

```

configurations/gdscript-configuration.json
```diff
@@ -1,36 +1,93 @@
...
-	}
+        "offSide": true,
+        "markers": {
+            "start": "^\\s*#\\s*region\\b",
+            "end": "^\\s*#\\s*endregion\\b"
+        }
+    },
+    "onEnterRules": [
+        {
+            "beforeText": "^\\s*(?:func|for|if|elif|else|while|match).*?:\\s*$",
```
Calinou, line 80: Is this also needed for `class`, `static func` and `enum` (as done above)?

was-games: What I noticed when looking into your question, is that I don't need to override the behavior described in the `indentationRules` (more specifically, I think that the `indentationRules` takes priority over `onEnterRules`). Basically, only the second rule is needed for the #337 to be solved.

Calinou:
Just to make sure, is this good to merge now?

was-games:
@Calinou Please see the behavior on the gif below.
337_fix
Moving the line after the if statement down works correctly.
Pasting after the if statement works correctly.

Adding indentation by just pressing Enter works as before.
I also tested other statements that have indentation rules.

The problem I'm having with this is that I can't explain why it fixes the issue, so if someone is willing to test this config as well, as I have only one environment to test on, it would be better.

DaelonSuzuka:
This is a big improvement over the previous behavior.

Calinou (approved):
Thanks!

PR #348, opened 2022-04-18
Add missing NodePath-using functions

DaelonSuzuka:
I missed a couple of functions that should have the special highlighting for NodePath parameters.


Calinou (approved):
Thanks!

PR #353, opened 2022-04-27
Fix keywords being incorrectly highlighted as function calls

DaelonSuzuka:
Some instances of keywords followed by an open paren were being incorrectly highlighted as a function call. 

I also added a `.prettierrc.json` config file to the syntaxes directory. I would really like to reach the point where the entire project has automatic formatting preconfigured, maybe including precommit hooks, but I'm starting with the folder I've touched the most.

(Also: hopefully I got all the trailing newlines correct this time...)

<details>
<summary>Screenshot</summary>

Code_k7imWQpXnI

</details>

### DIFF
--- a/syntaxes/.prettierrc.json
+++ b/syntaxes/.prettierrc.json
@@ -1,0 +1,4 @@
+{
+	"useTabs": false,
+	"tabWidth": 4
+}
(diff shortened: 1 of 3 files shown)


Calinou (approved):
Thanks!

PR #354, opened 2022-05-01
feature: enhanced comments to allow for @annotation for better readability

kaisershahid:
the following comment syntax is now supported:

- `# @annotation description`
- `# @annotation {type[,...]} description`
- `# @annotation id {type[,...]} description`

which looks like this on my machine:

Screen Shot 2022-05-01 at 5 57 55 PM

i find this very helpful for documenting signals and arguments whose type can't be explicitly defined (due to potential circular dependency)

### DIFF
--- a/syntaxes/GDScript.tmLanguage.json
+++ b/syntaxes/GDScript.tmLanguage.json
@@ -104,9 +104,50 @@
             "captures": {
                 "1": {
                     "name": "punctuation.definition.comment.number-sign.gdscript"
+                },
+                "2": {
+                    "name": "punctuation.definition.comment.number-sign.gdscript",
+                    "patterns": [
+                        {
+                            "_comment": "# @annotation id {type,...} description",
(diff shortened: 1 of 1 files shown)


atirut-w:
I think Godot 4 already have a docstring format for the builtin GDScript doc generator. Maybe we should follow that instead?

Calinou:
I wouldn't add syntax highlighting for nonstandard syntax of this kind, especially if it can be confused with annotations in Godot 4.0.

kaisershahid:
ahh, i wasn't aware. i'll check out the 4.0 syntax and follow up with changes for that

atirut-w:
> also, i want to clarify there's no type-hints for signal definitions, where i personally find some annotation necessary and clearly visible

Isn't GDScript's existing syntax clear enough?
```gdscript
func half(num: float) -> float:
    return num / 2
```

Calinou:
@atirut-w OP is referring to custom signals declared with `signal`. These only support type hints in 4.0, not 3.x.

atirut-w:
Ah, I see

Calinou:
Closing per the above comments. Thanks for the contribution nonetheless!

PR #357, opened 2022-05-06
Add context menu options to copy resource path

DaelonSuzuka:
Added an option to copy a file's resource path to the system clipboard, accessible at:
- file explorer's context menu
- editor tab title context menu
- the command palette(targeting the file in the last active editor)

Additionally, I fixed the technique used to resolve the `project.godot` config file. Previously, it assumed that the file was located in the root of the VSCode workspace, but that's not guaranteed to be the case.

I had to add a dependency to do this(klaw, for recursively directory searching), but I'm not 100% clear on how package-lock.json works and what I'm supposed to commit. I also don't know how the releases work, so hopefully @Calinou will let me know if I broke something.

Related future work:

- add a command palette command to grab the resource path of the currently edited/focused file
- improve overall extension startup time (it's seriously awful)
- cache the location of `project.godot` so we don't have to find it every boot

### DIFF
--- a/src/godot-tools.ts
+++ b/src/godot-tools.ts
@@ -70,12 +70,12 @@ export class GodotTools {
 
 		return new Promise<void>((resolve, reject) => {
 			let valid = false;
-			if (this.workspace_dir) {
+			if (this.project_dir) {
 				let cfg = this.project_file;
 				valid = (fs.existsSync(cfg) && fs.statSync(cfg).isFile());
 			}
 			if (valid) {
-				this.run_editor(`--path "${this.workspace_dir}" ${params}`).then(() => resolve()).catch(err => {
+				this.run_editor(`--path "${this.project_dir}" ${params}`).then(() => resolve()).catch(err => {
 					reject(err);
 				});
 			} else {
--- a/src/godot-tools.ts
+++ b/src/godot-tools.ts
@@ -11,12 +11,12 @@ export class GodotTools {
 	private reconnection_attempts = 0;
 	private context: vscode.ExtensionContext;
 	private client: GDScriptLanguageClient = null;
-    // deprecated, need to replace with "vscode.workspace.workspaceFolders", but
-    // that's an array and not a single value
+	// deprecated, need to replace with "vscode.workspace.workspaceFolders", but
+	// that's an array and not a single value
 	private workspace_dir = vscode.workspace.rootPath;
 	private project_file_name = "project.godot";
 	private project_file = "";
-    private project_dir = ""
+	private project_dir = ""
 	private connection_status: vscode.StatusBarItem = null;
 
 	constructor(p_context: vscode.ExtensionContext) {
@@ -49,15 +49,15 @@ export class GodotTools {
 		this.connection_status.command = "godot-tool.check_status";
 		this.connection_status.show();
 
-        // TODO: maybe cache this result somehow
-        const klaw = require('klaw');
-        klaw(this.workspace_dir)
-        .on('data', item => {
-            if (path.basename(item.path) == this.project_file_name) {
-                this.project_dir = path.dirname(item.path);
-                this.project_file = item.path;
-            }
-        });
+		// TODO: maybe cache this result somehow
+		const klaw = require('klaw');
+		klaw(this.workspace_dir)
+			.on('data', item => {
+				if (path.basename(item.path) == this.project_file_name) {
+					this.project_dir = path.dirname(item.path);
+					this.project_file = item.path;
+				}
+			});
 
 		this.reconnection_attempts = 0;
 		this.client.connect_to_server();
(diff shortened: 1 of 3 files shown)


Calinou:
> I had to add a dependency to do this(klaw, for recursively directory searching), but I'm not 100% clear on how package-lock.json works and what I'm supposed to commit.

Changes to both `package.json` and `package-lock.json` must be committed to version control, to ensure package installation is deterministic across machines.

DaelonSuzuka:
>Changes to both package.json and package-lock.json must be committed to version control, to ensure package installation is deterministic across machines.

I was afraid that was the answer. Since the very first time I cloned this project, doing the initial `npm install .` has dramatically changed package-lock.json, suggesting to me that it was already radically out of date(and has been for months, maybe). 

I had managed to ignore this until now by not touching the dependencies, assuming that if it already worked it would continue working.

Would it be right for me to add the most recent version to this PR? It still seems to build without it.

>Does this PR make the extension slower to start in large projects (with 1,000+ files)? If so, we need to address this before it can be merged.

Not in my testing.

Calinou:
> I was afraid that was the answer. Since the very first time I cloned this project, doing the initial npm install . has dramatically changed package-lock.json, suggesting to me that it was already radically out of date(and has been for months, maybe).
>
> I had managed to ignore this until now by not touching the dependencies, assuming that if it already worked it would continue working.

Indeed, I intentionally kept the lockfile version to 1 for now (npm has upgraded to v2 recently). I'm not sure if VS Code supports installing extensions with v2 lockfiles – it probably doesn't matter from the VS Code side anyway.

DaelonSuzuka:
That makes a lot more sense, thank you for the explanation. 

I'll get that committed in the next hour.

DaelonSuzuka:
>intentionally kept the lockfile version to 1 for now (npm has upgraded to v2 recently).

Just this information was enough for me to finally understand what I was looking at. Thanks again.

src/godot-tools.ts
```diff
@@ -11,8 +11,12 @@ export class GodotTools {
 	private reconnection_attempts = 0;
 	private context: vscode.ExtensionContext;
 	private client: GDScriptLanguageClient = null;
+    // deprecated, need to replace with "vscode.workspace.workspaceFolders", but
+    // that's an array and not a single value
```
Calinou, line 15:
```suggestion
	// deprecated, need to replace with "vscode.workspace.workspaceFolders", but
	// that's an array and not a single value
```

src/godot-tools.ts
```diff
@@ -11,8 +11,12 @@ export class GodotTools {
...
 	private workspace_dir = vscode.workspace.rootPath;
-	private project_file = "project.godot";
+	private project_file_name = "project.godot";
+	private project_file = "";
+    private project_dir = ""
```
Calinou, line 19:
```suggestion
	private project_dir = ""
```

src/godot-tools.ts
```diff
@@ -38,11 +42,23 @@ export class GodotTools {
...
 
+        // TODO: maybe cache this result somehow
+        const klaw = require('klaw');
+        klaw(this.workspace_dir)
+        .on('data', item => {
+            if (path.basename(item.path) == this.project_file_name) {
+                this.project_dir = path.dirname(item.path);
+                this.project_file = item.path;
+            }
+        });
```
Calinou, line 60:
```suggestion
		// TODO: maybe cache this result somehow
		const klaw = require('klaw');
		klaw(this.workspace_dir)
			.on('data', item => {
				if (path.basename(item.path) == this.project_file_name) {
					this.project_dir = path.dirname(item.path);
					this.project_file = item.path;
				}
			});
```

src/godot-tools.ts
```diff
@@ -69,6 +85,22 @@ export class GodotTools {
...
+        if (!uri) {
+            uri = vscode.window.activeTextEditor.document.uri
+        }
+
+        var relative_path = path.normalize(path.relative(this.project_dir, uri.fsPath));
+        relative_path = relative_path.split(path.sep).join(path.posix.sep);
+        relative_path = 'res://' + relative_path;
+
+        vscode.env.clipboard.writeText(relative_path);
+    }
```
Calinou, line 102:
```suggestion
		if (!this.project_dir) {
			return;
		}
        
		if (!uri) {
			uri = vscode.window.activeTextEditor.document.uri
		}

		let relative_path = path.normalize(path.relative(this.project_dir, uri.fsPath));
		relative_path = relative_path.split(path.sep).join(path.posix.sep);
		relative_path = 'res://' + relative_path;

		vscode.env.clipboard.writeText(relative_path);
    }
```

DaelonSuzuka:
> > > improve overall extension startup time (it's seriously awful)

> >    Does this PR make the extension slower to start in large projects (with 1,000+ files)? If so, we need to address this before it can be merged.

>Not in my testing.

Sorry, I meant to expand on this earlier. 

No, this PR doesn't worsen the extension's performance. It does queue some asynchronous filesystem activity(finding `project.godot` that must be completed before it's possible to produce resource paths, but it has yet to take more than a second. It appears to always finish the work I queue very shortly (<100ms?) after the rest of the extension appears ready(the status bar button appears, etc). 

General extension performance is pretty bad, though. One person I know has reported a startup time of **_1400ms_**, as measured by the tooltip in the Extensions sidebar. I suspect there's some O(n^2) action happening in the Language Client startup, or maybe it's accidentally blocking when it starts up the TCP socket, but I'm sure I'll find the answer when I start working on LSP improvements.

(note to self: learn how to profile extensions)

DaelonSuzuka:
Not trying to nag, but what's holding this up? 

I just want to make sure you're not waiting on something from me.

Calinou (approved):
Thanks!

Issue #356, opened 2022-05-03
Couldn't connect to the GDScript language server at 127.0.0.1:6008


oEjk:
### Godot version

v3.4.4.stable.arch_linux

### VS Code version

1.66.0

### Godot Tools VS Code extension version

v1.2.0

### System information

artix LXQt 

### Issue description

![screenshot-123e66d3](https://user-images.githubusercontent.com/104029652/166505459-e7edc990-03f6-4212-ba78-36cb4f47c175.png)

### Steps to reproduce

add the godot tool and ask it to work?

Calinou:
Related to https://github.com/godotengine/godot-vscode-plugin/issues/154.

oEjk:
> Just as a guess, are you running VSCode or Godot in a snap/flatpak/appimage? These containerized distribution tools sometimes have issues allowing inter-application communication like the LSP.

nope

> Failing that, are you sure Godot is actually running?

If the Godot Editor is running the issue resolves.

I misunderstood this...

Make sure to open the project in the Godot editor first. If you opened the editor after opening VS Code, you can click the Retry button in the bottom-right corner in VS Code.

as you could do this then close the editor. Perhaps I am slow, granted, or it may be mentioned the editor need to be running for the plugin to connect.

DaelonSuzuka:
Yes, the Godot Editor must be running for this extension to provide full functionality. This is the intended behavior, and not a bug.

raylu:
for anyone coming here from godot 4, the default port has changed to :6005
see #469

kepiz:
> for anyone coming here from godot 4, the default port has changed to :6005 see #469

how do you change the port?

sadeekgr:
the port needs to be changed to 6005 in the launch.json too

ElnuDev:
Shouldn't the default port in the extension be updated from 6008 to 6005 now that Godot 4 has been released for a while? This seems to be really bad user experience that one has to manually change the port configuration if one's using the latest version of Godot. A good compromise for supporting both Godot 3 and Godot 4 would be checking both ports. I don't think that this issue should be closed, as it makes the experience of the extension worse for new users.

atirut-w:
> I don't think that this issue should be closed, as it makes the experience of the extension worse for new users.

This issue was with 3.x, so it's an entirely different story. If my memory serves, there's an issue for 3-4 compat and definitely a PR for that.

PR #358, opened 2022-05-08
Mention that the Godot editor must be running in connection error message

Calinou:
See https://github.com/godotengine/godot-vscode-plugin/issues/356.

@DaelonSuzuka Could you take a look to make sure the message is well-worded?


DaelonSuzuka:
Sure thing.

Issue #362, opened 2022-05-15
Add ability to specify arbitrary arguments to Godot when launching the debugger


bitwes:
### Godot version

3.3 stable

### VS Code version

1.64.2

### Godot Tools VS Code extension version

1.3.0

### System information

macOS 10.15.7

### Problem statement

I discovered that other VSCode plugins can launch the debugger (provided by this plugin) using:
```
vscode.debug.startDebugging(...);
```
I would like to be able to send an arbitrary argument list when using this command.

I have a VSCode extension that runs GUT unit tests through VSCode (https://github.com/bitwes/gut-extension).  If I can pass arbitrary arguments to the debugger then I can run GUT through the debugger.  This would also give any other VSCode extension the ability to use the debugger for any purpose they want.

Here's an example of a GUT command:
```
 -s res://addons/gut/gut_cmdln.gd  -d  -gselect=test_bar.gd -ginner_class=TestBarClassTwo -gunit_test_name=test_bar_one
```
The `-s` option isn't currently supported by this plugin.  Also, there are over 15 CLI arguments available for GUT.  Being able to pass a string to launch the debugger would be great.

### Proposed solution

I was able to get a proof of concept working.  I added an `additional_options` parameter to this plugin (PR).  I then create a `DebugConfiguration` object and use it to launch the debugger with `vscode.debug.startDebugging(undefined, config);`.

Full Example
```
class GodotDebugConfiguration implements vscode.DebugConfiguration{
    public type = "godot";
    public name = "Debug Godot";
    public request = "launch";
    public project = "${workspaceFolder}";
    public port = 6007;
    public address = "127.0.0.1";
    public launch_game_instance = true;
    public launch_scene = false;
    public additional_options = "";
}

private runGutThroughDebugger(options:string = ''){
    let config = new GodotDebugConfiguration();
    config.additional_options = "-s \"res://addons/gut/gut_cmdln.gd\"";
    vscode.debug.startDebugging(undefined, config);
}
```

PR #363, opened 2022-05-15
Add "additional options" to launch debugger with

bitwes:
This PR implements #362.  

This allows other VSCode plugins to fully leverage the debugger interface that this plugin provides.  `additional_options` lets any other plugin specify whatever arguments it wants to send to the debugger.

### Usage Example
```
class GodotDebugConfiguration implements vscode.DebugConfiguration{
    public type = "godot";
    public name = "Debug Godot";
    public request = "launch";
    public project = "${workspaceFolder}";
    public port = 6007;
    public address = "127.0.0.1";
    public launch_game_instance = true;
    public launch_scene = false;
    public additional_options = "";
}

private runGutThroughDebugger(options:string = ''){
    let config = new GodotDebugConfiguration();
    config.additional_options = "-s \"res://addons/gut/gut_cmdln.gd\"";
    vscode.debug.startDebugging(undefined, config);
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -273,6 +273,11 @@
 								"type": "string",
 								"description": "Relative path from the godot.project file to a TSCN file. If launch_scene and launch_game_instance are true, and this file is defined, will launch the specified file instead of looking for an active TSCN file.",
 								"default": ""
+							},
+							"additional_options":{
+								"type":"string",
+								"description": "Additional command line arguments.",
+								"default":""
 							}
 						}
 					}
@@ -286,7 +291,8 @@
 						"port": 6007,
 						"address": "127.0.0.1",
 						"launch_game_instance": true,
-						"launch_scene": false
+						"launch_scene": false,
+						"additional_options":""
 					}
 				],
 				"configurationSnippets": [
@@ -300,7 +306,8 @@
 							"port": 6007,
 							"address": "127.0.0.1",
 							"launch_game_instance": true,
-							"launch_scene": false
+							"launch_scene": false,
+							"additional_options":""
 						}
 					}
 				]
--- a/src/debugger/debug_session.ts
+++ b/src/debugger/debug_session.ts
@@ -22,6 +22,7 @@ interface LaunchRequestArguments extends DebugProtocol.LaunchRequestArguments {
 	port: number;
 	project: string;
 	scene_file: string;
+	additional_options: string;
 }
 
 export class GodotDebugSession extends LoggingDebugSession {
@@ -235,6 +236,8 @@ export class GodotDebugSession extends LoggingDebugSession {
 			args.port,
 			args.launch_game_instance,
 			args.launch_scene,
+			args.scene_file,
+			args.additional_options,
 			get_configuration("scene_file_config", "") || args.scene_file,
 		]);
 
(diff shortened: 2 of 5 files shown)


Calinou:
cc @DaelonSuzuka

DaelonSuzuka:
This is just exposing functionality that the godot editor's debugger already has, right @bitwes? If that's the case, I don't see why this would be a problem. Improving interoperability is good.

I don't like to nitpick, but this commit _was_ more difficult to read than necessary because it included a bunch of pointless whitespace changes.

DaelonSuzuka:
Sorry @bitwes, this PR slipped my mind.

> I meant to remove it before submitting the PR. When I realized I hadn't I left it in and figured it would get addressed in the review. Should I remove it?

Well, I'll never tell someone *not* to clean up after themselves...

> You are correct that this is just allowing other extensions to pass arguments through to the debugger. No real functionality changes. I wasn't sure if the initialConfigurations and configurationSnippets in package.json should get the "additional_options" or not. Those are settings that a user of this extension would set, and this PR is adding functionality for other extensions. It doesn't hurt, it just may require more documentation to avoid confusion. I'd be happy to add some documentation or samples to the README.

Do you think an end user would ever want access to this feature from in VSCode? Adding a field to the extension's settings is easy, but then that introduces a bunch of extra complexity if a the user defines `additional_options` and then an extension wants to define it too. Does one take priority? Do they both just get shoved in?

For that matter, what would happen if two different extensions want to define different `additional_options`? Is there actually a conflict, or does the last one to set the field "win"? I doubt there's vscode extension authors lining up to tap into this feature, but we should think it through anyways.

bitwes:
>Do you think an end user would ever want access to this feature from in VSCode?

Maybe.  They could manually set options to run a standalone script (`extends SceneTree`) by setting options.  It would be tedious to use it that way, but it is something they couldn't do before.  I could see users wanting to specify options like `--low-dpi` or `--print-fps`.  This keeps you from having to handle all the options explicitly.  It's hard to say if this is worth the trouble though.  

There could be a collision between plugin and user specified `additional_options` if it is added to the settings.  I think you would want to use passed in options over settings.  That would allow plugins to specify the parameters they wanted without having user parameters clash.  This would have to be handled with a check.  If there are options passed in, use them, otherwise try to load them from the settings.  All settings should probably be treated this way.
```typescript
// in server_controller.ts

if (launch_instance) {
	let godot_path: string = utils.get_configuration("editor_path", "godot");
	let force_visible_collision_shapes = false;
	let force_visible_nav_mesh = false;

	let more_options:string = additional_options;
	if(!additional_options) {
		more_options =  utils.get_configuration("additional_options", "");
		force_visible_collision_shapes = utils.get_configuration("force_visible_collision_shapes", false);
		force_visible_nav_mesh = utils.get_configuration("force_visible_nav_mesh", false);
	}
	
	...
	executable_line += " " + more_options;
```

I don't think you would want to have one set of "additional options" for the user and another for plugins.  I could see this causing clashes in options which would be difficult for the user to figure out.

>For that matter, what would happen if two different extensions want to define different `additional_options`?

I don't think there can be any clashes between plugins since they are passing the values in (as opposed to changing setting values).

DaelonSuzuka (approved):

DaelonSuzuka:
> It's hard to say if this is worth the trouble though.

I think we're all going with "not worth".

> I don't think there can be any clashes between plugins since they are passing the values in (as opposed to changing setting values).

Sounds good to me.

@bitwes Thank you for your patience, both in explaining this feature to me in very small words, and for tolerating the delay in this PR getting reviewed.

Issue #370, opened 2022-05-25
Version 1.3.0 broke debugging


Exerionius:
### Godot version

v3.4.4.stable.official [419e713a2]

### VS Code version

1.67.2

### Godot Tools VS Code extension version

1.3.0

### System information

Windows 10

### Issue description

Debugging doesn't work in the latest version of the extension (1.3.0). When I press F5 it takes several seconds to start the game (before it was almost instant), and none of the break points work.

The configuration is correct, I triple-checked it.

Rolling back the extension to version 1.2.0 solves the problem completely.

Version 1.1.3 also works fine.

### Steps to reproduce

1. Have a default setup of Godot and VS Code + extension:

- Godot v3.4.4
- VS Code Extension v1.3.0
- Godot language port 6008
- Godot debug port 6007
- VS Code launch.json:

``` JSON
{
	"version": "0.2.0",
	"configurations": [
		{
			"name": "Godot",
			"type": "godot",
			"request": "launch",
			"project": "${workspaceFolder}",
			"port": 6007,
			"address": "127.0.0.1",
			"launch_game_instance": true,
			"launch_scene": false
		}
	]
}
```

2. Setup a couple of break-points
3. Hit F5
4. None of the break-points work
5. Roll-back extension version to 1.2.0
6. Reload VS Code
7. Hit F5
8. Break-points work

Exerionius:
1.2.0 command line:

`"D:\\Programs\\Godot\\Godot_3.4.4\\Godot_v3.4.4-stable_win64.exe"  --path "D:\GodotProjects\GooniesReloaded" --remote-debug 127.0.0.1:6007 --breakpoints res://state_machine/state_machine.gd:15,res://state_machine/state_machine.gd:19`

1.3.0 command line: 

`"D:\\Programs\\Godot\\Godot_3.4.4\\Godot_v3.4.4-stable_win64.exe"  --path "D:\GodotProjects\GooniesReloaded" --remote-debug 127.0.0.1:6007"""" --breakpoints res://state_machine/state_machine.gd:15,res://state_machine/state_machine.gd:19`

Looks like adding visible collision shapes and nav meshes broke the resulting execution command here: https://github.com/godotengine/godot-vscode-plugin/compare/1.2.0...1.3.0#diff-99b11a9a69c4124545c0a47d015c6bb5c5e665413cdcf52bbda3163fc72b273aR111

Calinou:
[...]

PR #371, opened 2022-05-25
Fix regression in launching debugger

DaelonSuzuka:
A previous commit assembled the `executable_line` variable incorrectly, and added redundant local variables.

This is a fix for #370. I can confirm that in `1.3,0`, launching is slower and breakpoints don't stop. This change fixes that regression.

However, on `1.2.0`, `1.3.0`, _and_ this patch, continuing or stepping over/in/out of a stopped breakpoint doesn't seem to do anything. I have to admit I've never used the VSCode debugger until now, so I'm not sure if this is a known issue, a problem with my configuration, or another regression.

### DIFF
--- a/src/debugger/server_controller.ts
+++ b/src/debugger/server_controller.ts
@@ -103,12 +103,12 @@ export class ServerController {
 
 			let executable_line = `"${godot_path}" --path "${project_path}" --remote-debug ${address}:${port}`;
 
-            if (force_visible_collision_shapes) {
-                executable_line += ` --debug-collisions`;
-            }
-            if (force_visible_nav_mesh) {
-                executable_line += ` --debug-navigation`;
-            }
+		if (force_visible_collision_shapes) {
+			executable_line += " --debug-collisions";
+		}
+		if (force_visible_nav_mesh) {
+			executable_line += " --debug-navigation";
+		}
 			if (launch_scene) {
 				let filename = "";
 				if (scene_file) {
--- a/src/debugger/server_controller.ts
+++ b/src/debugger/server_controller.ts
@@ -103,12 +103,12 @@ export class ServerController {
 
 			let executable_line = `"${godot_path}" --path "${project_path}" --remote-debug ${address}:${port}`;
 
-		if (force_visible_collision_shapes) {
-			executable_line += " --debug-collisions";
-		}
-		if (force_visible_nav_mesh) {
-			executable_line += " --debug-navigation";
-		}
+			if (force_visible_collision_shapes) {
+				executable_line += " --debug-collisions";
+			}
+			if (force_visible_nav_mesh) {
+				executable_line += " --debug-navigation";
+			}
 			if (launch_scene) {
 				let filename = "";
 				if (scene_file) {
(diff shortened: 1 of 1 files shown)


Exerionius:
Stepping over/in works for me in `1.2.0`. Debugger also shows locals/members, call stack and scene tree correctly.
Editing variables at runtime also works in `1.2.0`.

src/debugger/server_controller.ts
```diff
@@ -100,15 +100,15 @@ export class ServerController {
...
-			let executable_line = `"${godot_path}" --path "${project_path}" --remote-debug ${address}:${port}""${visible_collision_shapes_param}""${visible_nav_mesh_param}`;
+
+			let executable_line = `"${godot_path}" --path "${project_path}" --remote-debug ${address}:${port}`;
+
+            if (force_visible_collision_shapes) {
+                executable_line += ` --debug-collisions`;
+            }
+            if (force_visible_nav_mesh) {
+                executable_line += ` --debug-navigation`;
+            }
```
Calinou, line 111:
```suggestion
			let executable_line = `"${godot_path}" --path "${project_path}" --remote-debug ${address}:${port}`;

		if (force_visible_collision_shapes) {
			executable_line += ` --debug-collisions`;
		}
		if (force_visible_nav_mesh) {
			executable_line += ` --debug-navigation`;
		}
```

src/debugger/server_controller.ts
```diff
@@ -100,15 +100,15 @@ export class ServerController {
...
-			let executable_line = `"${godot_path}" --path "${project_path}" --remote-debug ${address}:${port}""${visible_collision_shapes_param}""${visible_nav_mesh_param}`;
+
+			let executable_line = `"${godot_path}" --path "${project_path}" --remote-debug ${address}:${port}`;
+
+		if (force_visible_collision_shapes) {
+			executable_line += " --debug-collisions";
+		}
+		if (force_visible_nav_mesh) {
+			executable_line += " --debug-navigation";
+		}
```
Calinou, line 111:
```suggestion
			if (force_visible_collision_shapes) {
				executable_line += " --debug-collisions";
			}
			if (force_visible_nav_mesh) {
				executable_line += " --debug-navigation";
			}
```

DaelonSuzuka:
> Stepping over/in works for me in 1.2.0. Debugger also shows locals/members, call stack and scene tree correctly.
Editing variables at runtime also works in 1.2.0.

You're using Godot 3.4.4, and I was on 3.5beta5. When I switch back to 3.4.3(the one I still had downloaded), I can continue breakpoints and also see locals/members. 

I think this fixed it, but we really need somebody else to test it.

DaelonSuzuka:
@atirut-w what Godot version are you using?

DaelonSuzuka:
Okay, *part* of my problem was user error. Continue and stepping all appear to work on `v3.4.3.stable.official.242c05d12`.

Stepping in too much sometimes causes the whole thing to lock up, but I'm fairly confident that this either still my fault or an existing weirdness.

@Calinou I'm thinking the reported regression is fixed. Further investigation into the debugger interface is a very large task and I can't promise any sort of timetable. As long as @atirut-w (or you, or somebody other than me) tested this PR, then my recommendation is to merge and publish a patch version to get the most broken behavior away from users.

Calinou (approved):
Thanks!

PR #385, opened 2022-06-16
Fix overaggressive formatting when adding lines after `if` statement

ZachIsAGardner:
This fixes https://github.com/godotengine/godot-vscode-plugin/issues/337

With this change the behavior should be the same as Godot's built in script editor. Godot's behavior from what I understand leaves you alone, unless you are pressing **"Enter"** at the end of a line containing **":"**.

This article helped a lot https://code.visualstudio.com/api/language-extensions/language-configuration-guide


configurations/gdscript-configuration.json
```diff
@@ -77,10 +73,10 @@
...
-            "beforeText": "^\\s*$",
+            "beforeText": ":\\s*$",
             "action": {
-                "indent": "none"
+                "indent": "indent"
             }
         }
     ]
-}
+}
```
Calinou, line 82:
```suggestion
}

```

Calinou (approved):
Thanks!

PR #387, opened 2022-06-21
Implement ability to inspect in Debug Console

### LINKED ISSUE #381: Debug Console only shows variable's correct value after hovering it
### Godot version

3.4.4.stable

### VS Code version

1.68.0

v1.3.1

### System information

Windows 10

### Issue description

As you can see in the gif below the Debug Console doesn't reliably show the correct value of a variable. Only after highlighting a variable with the mouse will it show the correct value.

godot_dc

### Steps to reproduce

- Place breakpoint in VS Code
- Run project
- After hitting breakpoint type in a name of a variable in the Debug Console (it will be wrong)
- Highlight the variable with your mouse
- Typing the variable again will give you the correct value

ZachIsAGardner:
This solves: https://github.com/godotengine/godot-vscode-plugin/issues/381

This adds the ability to inspect data through the Debug Console whenever you hit a breakpoint.

In the gif below a breakpoint in **world.gd** is hit. world.gd has a **character.gd class instance**. This **character instance** is being inspected in various ways in the gif below. So now we can do stuff like:

- character -> list all values
- character.position.x -> get position
- character.hats[0].scale.x -> index into the first hat that character has and tell me it's x scale
- and everything else

All the types listed at the following should work: https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_basics.html

gd_debug_console_nispect_demo

**world.gd**
```
extends Node2D

var cycle: int = 1
var character: Character = null

func _ready():
	print("Ready!")

	character = Character.new("Zachary", Vector2(200, 200))
	character.sprite.scale = Vector2(0.5, 1)
	add_child(character)
	
func _process(_delta):
	cycle += 1

```

**character.gd**
```
class_name Character

extends Node2D

var some_bool: bool = false
var some_int: int = 2
var some_float: float = 5.5
var some_string: String = "Staring!"

var some_vector2: Vector2 = Vector2(5, 6)
var some_rect2: Rect2 = Rect2()
var some_vector3: Vector3 = Vector3(1, 2, 3)
var some_transform3d: Transform = Transform()
var some_plane: Plane = Plane(1,0,0,0)
var some_quat: Quat = Quat(0,0,0,1)
var some_aabb: AABB = AABB()
var some_basis: Basis = Basis()
var some_transform2d: Transform2D = Transform2D()

var some_color: Color = Color(1,1,1,1)
var some_node_path: NodePath = ""
var some_rid: RID = RID()
var some_object: Object = null

var some_array = ["1", 2, "three"]
var some_dictionary: Dictionary = { "some_key": "some value here", 1: "one" }

var hats: Array = []
var sprite: Sprite = null
var some_var: String = "Hello"
var direction: int = 1

func _init(name: String, position: Vector2):
	self.name = name
	self.position = position
	sprite = Sprite.new()
	sprite.name = "Sprite"
	sprite.texture = load("res://icon.png")
	add_child(sprite)
	
	for i in [-1,1]:
		var hat = Hat.new(name, Vector2(10 * i, -50))
		hats.append(hat)
		add_child(hat)
	some_object = hats[0] 

	pass
```

*Bugsquad edit: This closes #381.*

### DIFF
--- a/src/debugger/debug_session.ts
+++ b/src/debugger/debug_session.ts
@@ -231,16 +231,14 @@ export class GodotDebugSession extends LoggingDebugSession {
 			if (result.variable.name == "self") {
 				result.object_id = this.all_scopes
 					.find(x => x && x.name == "id" && x.scope_path == "@.member.self").value;
-			}
-			else if (key) {
+			} else if (key) {
 				var collection = path.split(".")[path.split(".").length - 1];
 				var collection_items = Array.from(root.value.entries())
 					.find(x => x && x[0].split("Members/").join("").split("Locals/").join("") == collection)[1];
 				result.object_id = collection_items.get
 					? collection_items.get(key)?.id
 					: collection_items[key]?.id;
-			}
-			else {
+			} else {
 				result.object_id = Array.from(root.value.entries())
 					.find(x => x && x[0].split("Members/").join("").split("Locals/").join("") == propertyName)[1].id;
 			}
@@ -275,8 +273,7 @@ export class GodotDebugSession extends LoggingDebugSession {
 					result: parsed_variable.value,
 					variablesReference: !this.is_variable_built_in_type(variable.variable) ? variable.index : 0
 				};
-			}
-			else {
+			} else {
 				response.success = false;
 				response.message = variable.error;
 			}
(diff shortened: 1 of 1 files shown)


coelhucas (commented):
I think you'll want to replace all usages of `var` for either `let` or `const` (if they're not changed), since `var` in JS uses the global scope and generally not recommended, as well as for standard matters.

Awesome addition!

DaelonSuzuka:
I believe I've correctly implemented this in #452. This was another good idea, thanks again @ZachIsAGardner!

PR #397, opened 2022-07-22
Update gdshader syntax

DaelonSuzuka:
Pull additional gdshader syntax improvements from @AlfishSoftware's work

The shader syntax looks good to me, but I'm currently trying to solicit feedback on the [gdshader example file](https://github.com/godotengine/godot-vscode-plugin/blob/master/syntaxes/examples/example.gdshader).


Calinou:
> but I'm currently trying to solicit feedback on the [gdshader example file](https://github.com/godotengine/godot-vscode-plugin/blob/master/syntaxes/examples/example.gdshader).

I think it's pretty complete now. That reminds me, I should figure out adding `.gdshader` syntax highlighting on GitHub (and updating the GDScript syntax highlighter used on GitHub too).

syntaxes/GDShader.tmLanguage.json
```diff
@@ -220,6 +378,14 @@
...
+            "name": "punctuation.separator.comma.gdshader",
+            "match": "[,]"
+        },
+        "operator": {
+            "name": "keyword.operator.gdshader",
+            "match": "\\<\\<\\=?|\\>\\>\\=?|[-+*/&|<>=!]\\=|\\&\\&|[|][|]|[-+~!*/%<>&^|=]"
         }
     }
-}
+}
```
Calinou, line 391:
```suggestion
}

```

DaelonSuzuka:
> I think it's pretty complete now. That reminds me, I should figure out adding .gdshader syntax highlighting on GitHub (and updating the GDScript syntax highlighter used on GitHub too).

Wait, we can do that?

DaelonSuzuka:
> I think it's pretty complete now.

I just don't know enough to tell. I have some

Calinou:
> Wait, we can do that?

Yes, GitHub uses a public repository as a base for its syntax highlighting. I can't find that repository right now though.

AlfishSoftware:
> I'm currently trying to solicit feedback on the [gdshader example file](https://github.com/godotengine/godot-vscode-plugin/blob/master/syntaxes/examples/example.gdshader)

I don't know if the purpose of that file is to be more like an "exhaustive" syntax test, but ...
in case it is, taking a look I didn't find tests involving:
- arrays (e.g. array constructors, array fields in structs (`A[2] b; A b[2];`), etc), both declaration and initialization
- precision modifiers (including in arrays and struct fields)
- `in`, `out` modifiers
- `const` definitions
- swizzling (e.g. `.argb`, `.stpq`, `.wzy`, etc); methods (I think `.length()` is the only one?)
- some constructs, like `switch` `case`/`default`, `break`/`continue`, `do` `while`, `discard`
- many syntax variations for numeric literals (uint, float)
- all possible operators (e.g. `&&` `-=`, etc)

I have already tested these things, so maybe my files are pointless, but, if you want, you can check [one of my test files](https://github.com/godotengine/godot-vscode-plugin/files/8782689/godot4docs.gdshader.txt) with things from the docs, to catch some you might be missing. Like I said before, I lost my other test file, which started [here](https://github.com/godotengine/godot-vscode-plugin/pull/360#issuecomment-1133815778). There's also one with some variations of float literal syntax [in this issue](https://github.com/godotengine/godot/issues/61448).

If the purpose of the file is something else, just ignore what I said.

DaelonSuzuka:
> I don't know if the purpose of that file is to be more like an "exhaustive" syntax test, but ...
> in case it is, taking a look I didn't find tests involving:

Yeah, that's exactly what it is.

> <feedback>

All fantastic suggestions, thanks for the write-up. I'll have to study your test files a bit and get come back to updating mine.

@Calinou I'll expand the gdshader test file later, so this PR is ready as far as I'm concerned.

Calinou (approved):
Thanks a ton!

Issue #396, opened 2022-07-22
`func` keyword not highlighted unless function name and parentheseses are added


atirut-w:
### Godot version

N/A

### VS Code version

1.69.1

### Godot Tools VS Code extension version

1.3.1

### System information

N/A

### Issue description

Title describes the issue.

Screenshots:
![image](https://user-images.githubusercontent.com/25323231/180376899-256489fc-1c07-493e-b013-933adc4e38cb.png)
![image](https://user-images.githubusercontent.com/25323231/180376926-c08e1a06-cc18-4533-941e-42439c709ef4.png)

### Steps to reproduce

N/A

Calinou:
I think this was intentionally done in the GDScript syntax highlighter refactor. I'm not sure if it's a good idea as this isn't consistent with most other languages' syntax highlighters, though.

DaelonSuzuka:
PR is up.

This was unintended, probably missed when I rebuilt the function signature highlighting. The fix here is deceptively complicated, because adding `func` back to the list of keywords breaks the function signature detection, so an additional rule just for this keyword needed to be added.

@atirut-w Thank you very much for catching and reporting this.

Calinou:
Fixed by https://github.com/godotengine/godot-vscode-plugin/pull/398.

PR #398, opened 2022-07-22
Fix func keyword highlighting

DaelonSuzuka:
This is a fix for #396


Calinou (approved):
Thanks again for your continued efforts :slightly_smiling_face:

PR #402, opened 2022-08-05
Fix OS singleton being incorrectly highlighted as a constant

DaelonSuzuka:
Code_55p7kAtleh


Calinou:
OS is not a primitive type, so it shouldn't be highlighted this way. The issue is likely elsewhere.

DaelonSuzuka:
All the other built-in singletons are PascalCase so they're green like classes.

image

Calinou:
My guess is that the `const` highlight is kicking in as the identifier's name is all-uppercase.

DaelonSuzuka:
Yeah, that's exactly what's happening. OS doesn't trigger the same rule as the other singletons because it doesn't have any lower case letters. This change is just bringing OS in line with the other singletons.

Calinou (approved):
Thanks!

PR #403, opened 2022-08-05
Add highlighting support for %Unique nodes in NodePaths

DaelonSuzuka:
This PR adds support for the new Scene Unique Node syntax in NodePaths.

The highlighting works in `$"Shorthand"` and in string arguments to common NodePath-using functions.

image

I added a bonus feature that highlights invalid Unique nodes inside bare shorthands as red, eg: `$%Unique` or `$Child/%Unique`, because as far as I could tell, that's not supported syntax. In my opinion, this is not the job of the syntax highlighter, but a quick test shows me that the error displayed from the LSP is not exactly helpful if you don't already understand the feature:

image

My hope is that the highlighting will help the adoption of %Unique nodes, and I can come back later and either remove this highlighting or re-implement it in a more appropriate place.


syntaxes/GDScript.tmLanguage.json
```diff
@@ -749,4 +807,4 @@
             ]
         }
     }
-}
+}
```
Calinou, line 810:
```suggestion
}

```

Calinou (approved):
Thanks!

Issue #379, opened 2022-06-07
Missing keywords in GDScript grammar


AlfishSoftware:
### Godot version

v4.0.alpha8.official [cc3ed63af]

### VS Code version

1.67.2

### Godot Tools VS Code extension version

1.3.0

### System information

Kubuntu 22.04 x64

### Issue description

Some keywords are not being tokenized properly by GDScript grammar.
`await` - should be recognized as (control?) keyword
`super` - should be a keyword, since `super.etc` and `super(...)` have different semantics and it appears in tokenizer
`trait`, `namespace` - not mentioned in the [docs](https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_basics.html#keywords), but they appear in the [tokenizer in master](https://github.com/godotengine/godot/blob/master/modules/gdscript/gdscript_tokenizer.cpp) (reserved words?)
`in` - it's always being marked as `keyword.control.gdscript`; but when it's not part of a `for` construct, the `in` [operator](https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_basics.html#operators) should be `keyword.operator.wordlike.gdscript`

### Steps to reproduce

Test syntax coloring in a file like this:
```gdscript
extends Node
func f():
    await $Button.button_up
    super()
    super.some_function()
    for i in range(1): # `in` is a control keyword
        print(i in range(1)) # `in` is an operator keyword
```
To test `in` keyword on themes that don't distinguish control keywords, you can use the VSCode command "Developer: Inspect Editor Tokens and Scopes".

AlfishSoftware:
Strangely, `get` and `set` are also not being recognized as keywords like they appear to be in [this example in docs](https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_basics.html#properties):
```gdscript
signal changed(new_value)
var warns_when_changed = "some value":
    get:
        return warns_when_changed
    set(value):
        changed.emit(value)
        warns_when_changed = value
```
The editor in Godot 4 is not colorizing them as well, and they don't appear in the keyword list in the tokenizer.
Are they contextual keywords? Even if so, there should be a way to support them in the grammar as well.

atirut-w:
> ```gdscript
> signal changed(new_value)
> var warns_when_changed = "some value":
>     get:
>         return warns_when_changed
>     set(value):
>         changed.emit(value)
>         warns_when_changed = value
> ```

[...]

PR #407, opened 2022-08-13
Fix various highlighting errors

DaelonSuzuka:
This PR will have a collection of syntax highlighting improvements, starting with issues outlined in #379, 

- [x] `in` keyword

image

- [x] `await` and `super`

Code_rfH2Tv7ExD

- [x] `get` and `set` (harder)

Code_fffonCRmJX

- [x] `@annotation` syntax

I'm not sure if the list of "decorator" keywords is exhaustive, but the basic mechanism definitely works.

image

- [x] investigate other new Godot 4 features
- [x] lambdas

Code_uHjnckmxFp

- [x] type arrays

image

- [x] `trait` and `namespace`


DaelonSuzuka:
@Calinou I couldn't find any other new GDScript features that were missing syntax highlighting.

Let me know if you hate any of my colorization choices. I was torn on giving `@annotations` a more exotic color, but I actually think it looks fine like this.

Calinou:
> Let me know if you hate any of my colorization choices. I was torn on giving @annotations a more exotic color, but I actually think it looks fine like this.

Does VS Code have a standard annotation color (e.g. the one used in Python, or for preprocessor keywords in C++)?

syntaxes/GDScript.tmLanguage.json
```diff
@@ -807,4 +894,4 @@
             ]
         }
     }
-}
+}
```
Calinou, line 897:
```suggestion
}

```

syntaxes/examples/gdscript2.gd
```diff
@@ -0,0 +1,55 @@
+extends Node
```
Calinou, line 1: I think it'd be good idea to submit this code as a GDScript integration test in godotengine/godot :slightly_smiling_face: 

I can look into doing that.

Calinou (approved):
Thanks!

We can change the annotation color in a future PR.

syntaxes/examples/gdscript2.gd
DaelonSuzuka: Is there an ergonomic way to automate a highlighting test? I haven't been able to think of one that wouldn't be a massive pain to define the success/failure conditions.

I think finding a way to share test files with the engine is a great idea, though. Having files that we know exercise all of the compiler's features would really increase confidence that we're covering all the possible syntax.

DaelonSuzuka:
> Does VS Code have a standard annotation color (e.g. the one used in Python, or for preprocessor keywords in C/C++)?

The current annotation color is the same as python's decorators. Godot's script editor colors the annotations like keywords, but I thought that didn't look quite as good here. C/C++ preprocessor things are mostly just blue like keywords, as far as I remember. I know rust has some complicated syntax, I can see what they do for meta-informational markings like this. C# and C++ might also have something similar.

Issue #280, opened 2021-06-19
Unique global class "..." already exists at path: ... (there's only one class named like this)


ErnestoBorio:
As you can see in the screenshot, the file where class `ZXprite` is defined is `res://src/ZXprite.gd`, so this error seems to be complaining about this class itself although it's not defined anywhere else.  
It's connected with the godot language server and I don't get this error in godot's built-in editor.

![Screen Shot 2021-06-19 at 18 39 19](https://user-images.githubusercontent.com/1771719/122656062-d3147980-d12d-11eb-9ae0-38d4f799dde2.png)

Calinou:
Do you still get this error if you restart both Godot and VS Code? Or just VS Code?

ErnestoBorio:
Sorry, don't remember how I fixed it, but I did. Whenever I resume this project I'll take a look and close this issue.
Thanks!

bryab:
I had this issue and it was resolved by restarting Godot Editor (not restarting VS Code)

nickgirga:
I've found what was causing this issue for me and can reproduce and solve the issue. The reason it was happening for me was because I was using a symlink to a directory on an external drive to access my project in VSCode (`~/Projects` -> `/run/media/nick/LinuxStorage/Projects`). I noticed when I opened the file directly using the full path, it didn't show the error. So, I closed the folder and reopened it in VSCode using the full path on the external drive and the errors went away even in the project explorer and the problems panel. To see if that really was the issue, I closed the folder and reopened VSCode using the symlink, and sure enough, it gave me the error again. Closed and reopened VSCode with the full path and the error went away completely again. For me, I'm certain it was the fact that I was using symlinks/shortcuts (and possibly the external drive played a role).

Muha0644:
I am having the same issue, i am not using any symlinks. My project is in my home directory and vscode. The error is only present in VSC and not in Godot. Every class I define is flagged as a duplicate of itself.

raietracer:
I'm trying to figure out a workaround for this issue.  I'm having it as well.  It occurs for me in a project that is a repository clone.  I've tried opening all of the given suggestions and checked for symlinks but to my knowledge there isn't one.

DaelonSuzuka:
As far as I can tell, it just does this sometimes. I'm fairly confident that the error is coming from the language server and can't really be fixed in this extension.

I could probably add an option to suppress the error locally if anybody is interested in a band-aid fix.

[...]

Issue #266, opened 2021-05-03
'The identifier X isn't declared in the current scope' in inherited classes


PLyczkowski:
I get the 'The identifier X isn't declared in the current scope' even though it (the variable) is defined in the class the current class is inheriting from. Using Godot 3.3. stable

PLyczkowski:
Hmm, switching these lines twice seems to have removed the error?
![image](https://user-images.githubusercontent.com/7307968/116863158-2d9d5900-ac06-11eb-83c3-527fe72c1e00.png)

Issue #189, opened 2020-06-03
Create new built in function adds parentheses


TheColorRed:
When creating a function such as `_ready`, `_process`, etc., opening and closing parentheses is created after the colon:

![ExtraParen](https://user-images.githubusercontent.com/1571806/83687529-7f320380-a5b1-11ea-9d9d-f5a264401637.gif)

Calinou:
Which other VS Code extensions enabled? Maybe one of them could interfere with autocompletion.

TheColorRed:
There shouldn't be any that  interfere, these are the only ones I have enabled:

![image](https://user-images.githubusercontent.com/1571806/83760047-01fea100-a63a-11ea-98b3-d32306076225.png)

TheColorRed:
After testing more out, they each seem to work differently.

```
func _get(property: String):
func _get_configuration_warning() -> String:
func _draw() -> void:()
func _process(delta: float) -> void:(
func _unhandled_input(event: InputEvent) -> void:(
```
* Some have return types
* Some don't have return types
* Some are correct
* Some have parentheses
* Some don't have parentheses
* Some have an opening parentheses

Calinou:
I think `_get()` doesn't have a return type because it can be anything. GDScript doesn't have an `any` return type yet (and doesn't allow to use `Variant` as a return type), so it has to be ommitted for now.

TheColorRed:
Okay, I have removed that from the list of tested items.

Razoric480:
`func _get_configuration_warning() -> String:` adds parenthesis at the end for me, so it's also not correct. Though the other 3 are as stated.

Razoric480:
I've finally investigated this eons later and can confirm that it's an issue with the LSP.
```cpp
if ((item.kind == lsp::CompletionItemKind::Method || item.kind == lsp::CompletionItemKind::Function) && !item.label.ends_with("):")) {
		item.insertText = item.label + "(";
		if (symbol && symbol->children.empty()) {
			item.insertText += ")";
		}
	}
```
does not take into consideration instances where static typing is included with `->`. Changing it to
```cpp
if ((item.kind == lsp::CompletionItemKind::Method || item.kind == lsp::CompletionItemKind::Function) && !item.label.ends_with("):") && item.label.rfind("->") == -1) {
		item.insertText = item.label + "(";
		if (symbol && symbol->children.empty()) {
			item.insertText += ")";
		}
	}
```
removes the issue.

[...]

Issue #152, opened 2020-02-26
Close bracket is not generated when autocomplete


tienlocnguyen:
Only open bracket is generated, closed bracket is not
![Peek 2020-02-26 20-21](https://user-images.githubusercontent.com/6823008/75348842-21b55900-58d6-11ea-86da-2babbeca023a.gif)
Expect that close bracket will be generated too

colludium:
Any update on this please? Thanks.

atirut-w:
This is a very annoying issue but I've somehow forgotten about it. Maybe @DaelonSuzuka can fix this?

Calinou:
Related to https://github.com/godotengine/godot-vscode-plugin/issues/142.

DaelonSuzuka:
@atirut-w @Calinou Fixing this in the Language Server is a bit beyond my expertise at the moment. There are a number of other issues with both the Server(at least in 3.x) and the Client that I will eventually be addressing, but it's very complex system and that's going to take some time.

Luckily, I think there's a quick and dirty solution by just modifying the completions in the Client based on the context around the cursor/selection. If I have time this week, I'll try to put together a proof of concept.

DaelonSuzuka:
@geekley 

> So, please add at least an option to just not add any parentheses automatically.

This behavior is controlled by the Language Server, which is a component of the Godot Engine itself. This extension does not control the completions, so there's nothing I can do about it.

geekley:
@DaelonSuzuka is there an upstream issue which I can track then? So they add an option in the LSP?

geekley:
Found this, which I hope is the right issue, moved my comment there: https://github.com/godotengine/godot/issues/86488#issuecomment-2430017477

BilgiPasa:
I think this problem might be solved because I do not experience this issue in Godot 4.7 right now. I am using VSCodium 1.121 with godot-tools 2.7.1

HolonProduction:
Yeah 4.7 fixed this.

PR #410, opened 2022-08-17
Fix internal document link handling

DaelonSuzuka:
This PR adds a `DocumentLinkProvider` that creates internal hyperlinks for all `res://path/to/file.ext` strings in gdresource files.

Additionally, this provides a workaround for the Language Server's broken behavior of sending invalid file URIs in Godot 3. I tested the workaround with both the current broken 3.5stable LSP and the fixed version that I created a [PR](https://github.com/godotengine/godot/pull/64524) for. The same fix is already implemented in `master`, so that means this workaround should also work correctly with Godot 4.

https://user-images.githubusercontent.com/18042232/185079747-46fb067e-c7ff-4561-b31d-19a53793e43d.mp4

The technique used for this workaround could be useful for mitigating other defects in the Language Server, like fixing bad completions(#152, #189) or suppressing incorrect errors(#280, #266),


syntaxes/examples/Example.gd
```diff
@@ -0,0 +1,4 @@
+extends Node
+
+func _ready():
+	pass
```
Calinou, line 4:
```suggestion
	pass

```

Calinou (approved):
Thanks!

src/lsp/GDScriptLanguageClient.ts
```diff
@@ -97,6 +97,19 @@ export default class GDScriptLanguageClient extends LanguageClient {
...
+        // This should be forward-compatible, meaning that it will work
+        // with the current broken version, AND the fixed future version.
+        const match = JSON.stringify(message).match(/"target":"file:\/\/[^\/][^"]*"/);
+        if (match) {
+            for (let i = 0; i < message["result"].length; i++) {
+                const x = message["result"][i]["target"];
+                message["result"][i]["target"] = x.replace('file://', 'file:///');
+            }
+        }
+
```
Calinou, line 112: Use tabs for indentation.

Calinou (approved):
Thanks!

Issue #415, opened 2022-09-02
Grammar contains several invalid regex with lookbehind assertions that are not a fixed length


lildude:
### Godot version

N/A

### VS Code version

N/A

### Godot Tools VS Code extension version

master

### System information

N/A

### Issue description

👋 I'm the lead maintainer of the https://github.com/github/linguist library which is used for language detection and providing the syntax highlighting for languages on GitHub.com, and we use this grammar.

Our grammar compiler has found several problems with the grammars which I thought I'd let you know about:

- Invalid regex in grammar: `source.gdscript` (in `syntaxes/GDScript.tmLanguage.json`) contains a malformed regex (regex "`\b(?<=for\s[\w]*\s)(in)\b`": lookbehind assertion is not fixed length (at offset 18))
- Invalid regex in grammar: `source.gdshader` (in `syntaxes/GDShader.tmLanguage.json`) contains a malformed regex (regex "`(?<=[.]\s*)(?:[xyzw]{2,4}|[rgba]`...": lookbehind assertion is not fixed length (at offset 10))
- Invalid regex in grammar: `source.gdshader` (in `syntaxes/GDShader.tmLanguage.json`) contains a malformed regex (regex "`(?<=[.]\s*)[a-zA-Z_]\w`": lookbehind assertion is not fixed length (at offset 10))

### Steps to reproduce

View each of the regexes in turn in regex101 and note the errors:

- https://github.com/godotengine/godot-vscode-plugin/blob/e7674c12adef55d13d17ee66b057a3f1789a2102/syntaxes/GDScript.tmLanguage.json#L298

  https://regex101.com/r/TFZmub/1

- https://github.com/godotengine/godot-vscode-plugin/blob/e7674c12adef55d13d17ee66b057a3f1789a2102/syntaxes/GDShader.tmLanguage.json#L345

  https://regex101.com/r/JX03EY/1

- https://github.com/godotengine/godot-vscode-plugin/blob/e7674c12adef55d13d17ee66b057a3f1789a2102/syntaxes/GDShader.tmLanguage.json#L349

  https://regex101.com/r/2Zpu1I/1

Calinou:
cc @DaelonSuzuka 

PS: I didn't know this repository was used as a basis for Linguist.

lildude:
We've been using it for GDScript since 2020 thanks to https://github.com/github/linguist/pull/5000 😁

DaelonSuzuka:
[...]

PR #416, opened 2022-09-05
Fix errors in grammar syntax

DaelonSuzuka:
Per #415, this PR fixes some regexes that appeared to work in VSCode/onigumura but do not work in linguist using PCRE.

When this PR is done, I need to remember to post in github/linguist#3924 and let the linguist project know that we've fixed our grammars.

Additionally, @lildude has documented a process [here](https://github.com/godotengine/godot-vscode-plugin/issues/415#issuecomment-1237132824) that we can use to verify that these changes satisfy `linguist`'s requirements. It's worth investigating if this process or a derivative can be integrated into our CI, which will at least provide syntax validation of our grammar files.

All three of the errors are invalid uses of positive lookbehind using a non-fixed length.

Problem areas:

- https://github.com/godotengine/godot-vscode-plugin/blob/e7674c12adef55d13d17ee66b057a3f1789a2102/syntaxes/GDScript.tmLanguage.json#L298
- https://github.com/godotengine/godot-vscode-plugin/blob/e7674c12adef55d13d17ee66b057a3f1789a2102/syntaxes/GDShader.tmLanguage.json#L345
- https://github.com/godotengine/godot-vscode-plugin/blob/e7674c12adef55d13d17ee66b057a3f1789a2102/syntaxes/GDShader.tmLanguage.json#L349

I've already fixed the first problem, in the GDScript grammar, but unfortunately I've discovered that there was a regression sometime between `1.3.1` and now. The GDShader syntax doesn't even run, and there aren't any error messages about failing to load the grammar.

I'll have to find the cause of this regression before I can fix the two regex errors in GDShader.

### DIFF
--- a/syntaxes/GDScript.tmLanguage.json
+++ b/syntaxes/GDScript.tmLanguage.json
@@ -357,7 +357,7 @@
             },
             "patterns": [
                 {
-                    "match": "(:)\\s*([\\w]*)?",
+                    "match": "(:)\\s*([a-zA-Z_]\\w*)?",
                     "captures": {
                         "1": {
                             "name": "punctuation.separator.annotation.gdscript"
@@ -372,7 +372,7 @@
                     "name": "keyword.operator.assignment.gdscript"
                 },
                 {
-                    "match": "(setget)\\s+([\\w]*)(?:[,]\\s*([\\w]*))?",
+                    "match": "(setget)\\s+([a-zA-Z_]\\w*)(?:[,]\\s*([a-zA-Z_]\\w*))?",
                     "captures": {
                         "1": {
                             "name": "storage.type.const.gdscript"
@@ -693,7 +693,7 @@
                     "include": "#annotated_parameter"
                 },
                 {
-                    "match": "(?x)\n  ([[:alpha:]_]\\w*)\n    \\s* (?: (,) | (?=[)#\\n=]))\n",
+                    "match": "(?x)\n  ([a-zA-Z_]\\w*)\n    \\s* (?: (,) | (?=[)#\\n=]))\n",
                     "captures": {
                         "1": {
                             "name": "variable.parameter.function.language.gdscript"
@@ -731,7 +731,7 @@
             ]
         },
         "annotated_parameter": {
-            "begin": "(?x)\n  \\b\n  ([[:alpha:]_]\\w*) \\s* (:)\n",
+            "begin": "(?x)\n  \\b\n  ([a-zA-Z_]\\w*) \\s* (:)\n",
             "end": "(,)|(?=\\))",
             "beginCaptures": {
                 "1": {
@@ -796,7 +796,7 @@
         "function_call": {
             "name": "meta.function-call.gdscript",
             "comment": "Regular function call of the type \"name(args)\"",
-            "begin": "(?x)\n  \\b(?=\n    ([[:alpha:]_]\\w*) \\s* (\\()\n  )\n",
+            "begin": "(?x)\n  \\b(?=\n    ([a-zA-Z_]\\w*) \\s* (\\()\n  )\n",
             "end": "(\\))",
             "endCaptures": {
                 "1": {
@@ -823,7 +823,7 @@
                 {
                     "comment": "Some color schemas support meta.function-call.generic scope",
                     "name": "support.function.any-method.gdscript",
-                    "match": "(?x)\n  \\b ([[:alpha:]_]\\w*) \\b\n"
+                    "match": "(?x)\n  \\b ([a-zA-Z_]\\w*) \\b\n"
                 }
             ]
         },
@@ -842,7 +842,7 @@
                     "match": "(,)"
                 },
                 {
-                    "match": "\\b([[:alpha:]_]\\w*)\\s*(=)(?!=)",
+                    "match": "\\b([a-zA-Z_]\\w*)\\s*(=)(?!=)",
                     "captures": {
                         "1": {
                             "name": "variable.parameter.function-call.gdscript"
--- a/syntaxes/GDScript.tmLanguage.json
+++ b/syntaxes/GDScript.tmLanguage.json
@@ -286,8 +286,11 @@
         "in_keyword": {
             "patterns": [
                 {
-                    "match": "\\b(?:for\\s[\\w]*\\s)(in)\\b",
-                    "name": "keyword.control.gdscript"
+                    "match": "\\b(for)\\s+[a-zA-Z_]\\w*\\s+(in)\\b",
+                    "captures": {
+                        "1": { "name": "keyword.control.gdscript" },
+                        "2": { "name": "keyword.control.gdscript" }
+                    }
                 },
                 {
                     "match": "\\bin\\b",
--- a/syntaxes/GDScript.tmLanguage.json
+++ b/syntaxes/GDScript.tmLanguage.json
@@ -286,7 +286,7 @@
         "in_keyword": {
             "patterns": [
                 {
-                    "match": "\\b(?<=for\\s[\\w]*\\s)(in)\\b",
+                    "match": "\\b(?:for\\s[\\w]*\\s)(in)\\b",
                     "name": "keyword.control.gdscript"
                 },
                 {
--- a/syntaxes/GDScript.tmLanguage.json
+++ b/syntaxes/GDScript.tmLanguage.json
@@ -1,7 +1,5 @@
 {
-    "fileTypes": [
-        "gd"
-    ],
+    "fileTypes": ["gd"],
     "scopeName": "source.gdscript",
     "name": "GDScript",
     "patterns": [
@@ -419,7 +417,7 @@
                             "include": "#line_continuation"
                         },
                         {
-                            "match": "\\s*(\\-\\>)\\s*([a-zA-Z_][a-zA-Z_0-9]*)\\s*\\:",
+                            "match": "\\s*(\\-\\>)\\s*([a-zA-Z_]\\w*)\\s*\\:",
                             "captures": {
                                 "1": {},
                                 "2": {
@@ -451,7 +449,7 @@
                     "name": "storage.type.new.gdscript"
                 }
             },
-            "match": "\\b([a-zA-Z_][a-zA-Z_0-9]*).(new)\\("
+            "match": "\\b([a-zA-Z_]\\w*).(new)\\("
         },
         "class_is": {
             "captures": {
@@ -462,7 +460,7 @@
                     "name": "entity.name.type.class.gdscript"
                 }
             },
-            "match": "\\s+(is)\\s+([a-zA-Z_][a-zA-Z_0-9]*)"
+            "match": "\\s+(is)\\s+([a-zA-Z_]\\w*)"
         },
         "class_enum": {
             "captures": {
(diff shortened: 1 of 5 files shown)


DaelonSuzuka:
I have absolutely no idea how this happened, but there were a ton of regex special characters that were missing the second backslash(they need to be double escaped because reasons). I recopied the grammar from @alfishsoftware's extension, fixed our handful of differences, and it's working again.

I replaced the invalid lookbehinds and it seems to work. Next, I'll run `linguist`'s grammar compiler and see if we pass.

DaelonSuzuka:
I followed the steps to set up linguist, ran `script/bootstrap`, manually replaced the grammars it download with the fixed versions from this PR, ran `script/grammar-compiler add vendor/grammars/godot-vscode-plugin/`, and I believe this is a positive result:

```
daelon@isotope:~/linguist$ sudo script/grammar-compiler add vendor/grammars/godot-vscode-plugin/
latest: Pulling from linguist/grammar-compiler
Digest: sha256:203514bbf04b83d3594245338fc08097301e54139ccec20ab05861302c23ab64
Status: Image is up to date for linguist/grammar-compiler:latest
docker.io/linguist/grammar-compiler:latest
OK! added grammar source 'vendor/grammars/godot-vscode-plugin/'
        new scope: source.gdshader
        new scope: source.gdresource
        new scope: source.gdscript
daelon@isotope:~/linguist$ 
```

@lildude does this mean that your compiler is happy now?

@Calinou I think this is done, but it might be safer to wait for the thumbs-up from the `linguist` people before merging.

syntaxes/GDShader.tmLanguage.json
```diff
@@ -314,51 +314,51 @@
...
         "swizzling": {
             "name": "variable.other.property.gdshader",
-            "match": "(?<=[.]\\s*)(?:[xyzw]{2,4}|[rgba]{2,4}|[stpq]{2,4})\b"
+            "match": "(?:[.]\\s*)(?:[xyzw]{2,4}|[rgba]{2,4}|[stpq]{2,4})\\b"
         },
         "identifierField": {
             "name": "entity.name.variable.field.gdshader",
-            "match": "(?<=[.]\\s*)[a-zA-Z_]\\w*\b"
+            "match": "(?:[.]\\s*)[a-zA-Z_]\\w*\\b"
         },
```
AlfishSoftware, line 350: Just a minor comment on these. If lookbehind for the dot is replaced with making it part of the captured regex, then the editor will likely include the dot too as part of the field/property coloring. Which is fine if that's what is intended.
But in case this is undesired, you can put `[.]` in a capturing group and use something like:
```json
"captures": {
  "1": {
    "name": "punctuation.accessor.gdshader"
  }
},
```

DaelonSuzuka: I updated both rules, and added the example file you linked me to in a previous discussion.

I think my original version was actually broken, so thanks for the feedback. The new version actually works, as well as correctly tagging the period. I'd greatly appreciate it if you could proofread me again!

AlfishSoftware:
By the way, I didn't test it in VSCode (no time at the moment), so I don't know if the dot changes affect `.length()`. It might be the case where it picks `.length` as identifierField instead of `length()` as a identifierFunction as before, so you might want to double-check that length is colored correctly as a function, just in case.

DaelonSuzuka:
I'll definitely double check that, thanks for the tip.

DaelonSuzuka:
> It might be the case where it picks .length as identifierField instead of length() as a identifierFunction

It was doing exactly this. Fantastic catch, as usual.

syntaxes/GDShader.tmLanguage.json
```diff
@@ -314,51 +314,65 @@
...
+                },
+                "2": {
+                    "name": "variable.other.property.gdshader"
+                }
+            }
         },
         "identifierField": {
-            "name": "entity.name.variable.field.gdshader",
-            "match": "(?<=[.]\\s*)[a-zA-Z_]\\w*\b"
+            "match": "([.])\\s*([a-zA-Z_]+)\\w*\\b(?!\\()",
```
AlfishSoftware, line 355: Don't forget to capture the `\\w*`, like I said before; and remove the `+`:
"...`([a-zA-Z_]\\w*)`..."

Also, this case won't catch `.length ()` with a space (some people might use this kind of weird code style?). This can be solved with a `\\s*`, but I'm not familiar with linguist, so I don't know if you can use variable-length negative lookahead (presumably yes?). If you can't, maybe adding it right before the lookahead should be enough.

AlfishSoftware (changes requested):

AlfishSoftware:
Oh. I just realized review comments aren't published right away, it seems? Sorry about that, I guess you didn't see the first one before.

syntaxes/GDShader.tmLanguage.json
```diff
@@ -341,12 +341,26 @@
...
+                },
+                "2": {
+                    "name": "variable.other.property.gdshader"
+                }
+            }
         },
         "identifierField": {
-            "name": "entity.name.variable.field.gdshader",
-            "match": "(?:[.]\\s*)[a-zA-Z_]\\w*\\b"
+            "match": "([.])\\s*([a-zA-Z_]+)\\w*\\b",
```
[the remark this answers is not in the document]
DaelonSuzuka: I have no idea what I was thinking when I added those parens. 

My only defense is that I try to avoid using `\w` or other regex special characters because I have to use regexes across a bunch of different platforms, and I don't trust the implementations to have the same definitions, or myself to remember the differences.

syntaxes/GDShader.tmLanguage.json
DaelonSuzuka: >  not familiar with linguist, so I don't know if you can use variable-length negative lookahead (presumably yes?). If you can't, maybe adding it right before the lookahead should be enough.

I'll try and run both versions against their "compiler", just for the sake of education.

DaelonSuzuka:
> Oh. I just realized review comments aren't published right away, it seems? Sorry about that, I guess you didn't see the first one (marked outdated) before.

I replied to every comment I can see, and I marked the first one as 'resolved' after I (tried to) fix the relevant bits.

If there are other comments I didn't interact with, then I definitely blame github's rather obtuse UI.

syntaxes/GDShader.tmLanguage.json
DaelonSuzuka: Linguist's compiler did not complain about variable-length negative lookahead. Some reading I did earlier implied that despite being syntactically similar, lookahead and lookbehind are implemented dramatically differently in the regex state machine, with lookbehind being a much more complicated process. 

> `.length ()` with a space (some people might use this kind of weird code style?)

As much as I don't want to encourage this kind of behavior, this rule correctly supports whatever mangled whitespace somebody wants to insert. I'm gonna have to review my gdscript grammar and make sure it's similarly flexible.

Thanks for the regex review, I feel like I picked up some new tricks!

DaelonSuzuka:
@Calinou This PR is also done. I rebased it to master myself just to confirm that #419 doesn't conflict(both PRs touched the GDScript grammar).

syntaxes/GDShader.tmLanguage.json
AlfishSoftware: > I think the negative-lookahead also has stronger guarantees about not "consuming" what it matches against?

Yes! In VSCode grammar, lookaheads/behinds don't consume the input. At least the `\\(` has to be inside a lookahead; the `\\s*` could be outside it if needed (e.g. `\\b\\s*(?!\\()`), but since you say it'll likely work, then (B) is indeed better. (A) seems wrong, because `^` is start of line and also like I said `(` should not be consumed as part of the token.

syntaxes/GDScript.tmLanguage.json
```diff
@@ -286,7 +286,7 @@
         "in_keyword": {
             "patterns": [
                 {
-                    "match": "\\b(?<=for\\s[\\w]*\\s)(in)\\b",
+                    "match": "\\b(?:for\\s[\\w]*\\s)(in)\\b",
                     "name": "keyword.control.gdscript"
```
AlfishSoftware, line 290: I didn't test this, but I suspect that it will now consider the whole thing as a keyword.
E.g. `for i in array` then `for i in` is a keyword. It shouldn't mark the loop variable (`i` in this case) as a keyword.

The correct thing to do here is to join this with the `for` keyword so that one rule recognizes both keywords. You don't need to remove `for` from `control_flow` (`for` should still be a keyword even before `in` is typed) as long as this is before so it has higher priority; but change this rule to something along the lines of:
```json
{
  "match": "\\b(for)\\s+([a-zA-Z_]\\w*)\\s+(in)\\b",
  "captures": {
    "1": { "name": "keyword.control.gdscript" },
    "2": { "name": "variable.name.gdscript" },
    "3": { "name": "keyword.control.gdscript" }
  }
},
```
I changed `\w` (which matches `[a-zA-Z0-9_]` including numbers) since I believe identifiers can't start with a number (correct me if I'm wrong). It's probably worth checking this on other places in the code too. I'm assuming identifiers allow ASCII only.
I also allowed more than one space in the syntax.
Btw, using a tool like https://regex101.com/ while you write it really helps with understanding nuances in regexes and checking that it will really do what is expected, and prevent against corner cases.

DaelonSuzuka: I cannot believe that I missed this. You were correct, `i` was purple.

image

Your rule worked great. I removed the middle capture group and it's name block, because we don't actually use the `variable.name` tag for variables. Python doesn't tag variable names, they're just white, so that's the behavior I want to imitate.

> I changed \w (which matches [a-zA-Z0-9_] including numbers) since I believe identifiers can't start with a number (correct me if I'm wrong). It's probably worth checking this on other places in the code too. I'm assuming identifiers allow ASCII only.

Many other places used the full form `[a-zA-Z_][a-zA-Z_0-9]*`, but I just replaced them with `[a-zA-Z_]\\w*`. I wish we could give a block like this a custom name so we don't have to copy it everywhere. 

> Btw, using a tool like [regex101.com](https://regex101.com/) while you write it really helps with understanding nuances in regexes and checking that it will really do what is expected, and prevent against corner cases.

I started off doing this, but having to add all the extra escapes to move a rule over to vscode was too annoying so I stopped. I've thought about switching to yaml files, since the regex strings won't need escapes, but there was some problem with that too.

DaelonSuzuka:
I checked this latest commit with linguist again, and there's no warnings/errors.

I've decided that I'm definitely going to set up that grammar unit testing tool, but I don't have time right now.

syntaxes/GDScript.tmLanguage.json
AlfishSoftware: >> It's probably worth checking this on other places in the code too.

> Many other places used the full form `[a-zA-Z_][a-zA-Z_0-9]*`, but I just replaced them with `[a-zA-Z_]\\w*`.

I was actually referring to places capturing identifiers with just `\w+`, which would allow names starting with numbers, e.g. `1whatever`, `0e2f`, `0xFACADE`. I took a look and saw it in 2 places.
So unless this is on purpose for some reason, I would suggest:

[Line 360](https://github.com/godotengine/godot-vscode-plugin/pull/416/files#diff-2a31f168b0e81523889b916085b33a7c7a235830d79b2dc783b261579f88eb02R360):
```json
"match": "(:)\\s*([a-zA-Z_]\\w*)?",
```
[Line 375](https://github.com/godotengine/godot-vscode-plugin/pull/416/files#diff-2a31f168b0e81523889b916085b33a7c7a235830d79b2dc783b261579f88eb02R375):
```json
"match": "(setget)\\s+([a-zA-Z_]\\w*)?(?:[,]\\s*([a-zA-Z_]\\w*)?)?",
```

DaelonSuzuka: Another good catch, thanks. I found some other places where identifier names were being matched with `[[:alpha:]_]\\w*`, so I standardized all of them to `[a-zA-Z_]\\w*`.

DaelonSuzuka:
I checked the latest changes with linguist and it looks like they're good to go.

Calinou (approved):
Thanks!

AlfishSoftware:
Cool! Later I'll replicate those changes to gdshader (removing back-references) back in my own extension too, just in case.

aaronfranke:
Does this PR fix https://github.com/godotengine/godot-vscode-plugin/issues/415?

DaelonSuzuka:
> Does this PR fix https://github.com/godotengine/godot-vscode-plugin/issues/415?

It does, but there's been some delay in communicating with the Github LInguist project to make sure the original source of the reported issue is fixed. I intended to close all the related items together but then time got away from me.

Thanks for noticing my miss!

PR #422, opened 2022-09-09
Add "_unhandled_input" snippet

kemgoblin:
`uinput` seems godot to me, err I mean good


DaelonSuzuka (approved):

DaelonSuzuka:
I honestly forget that snippets are there, so thanks for contributing these!

Issue #435, opened 2022-11-03
Language server stuck "Initializing"


OsamaHeweitat:
### Godot version

3.5.1

### VS Code version

1.73.0

### Godot Tools VS Code extension version

1.3.1

### System information

Windows 10

### Issue description

Whenever VSCode is launched, regardless if through Godot or not, or whether it was launched before or after Godot, the language server status at the bottom right will be stuck as "Initializing", clicking on it will give the following popup text: "Connecting to the GDScript language server at 127.0.0.1:6007"

VSCode is actually not connecting to the language server and it isn't just a visual bug, as features such as auto-complete and debugging do not work. I tried every possible fix and nothing worked, I tried multiple versions of Godot and VSCode.

### Steps to reproduce

1. Launch VSCode.

OsamaHeweitat:
The issue doesn't occur when using versions of the plugin below 1.3.0, such as 1.2.0. Using 1.3.0 and 1.3.1 causes the issue to arise again.

Calinou:
cc @DaelonSuzuka

OsamaHeweitat:
Can confirm the issue still occurs for me using the latest `master` branch.

DaelonSuzuka:
Very interesting. I did a quick test and I can't replicate this on `1.3.0`, `1.3.1`, or `master` connecting to Godot `3.5` or `3.5.1` on Windows 10.

> popup text:  "Connecting to the GDScript language server at 127.0.0.1:6007"

@OsamaHeweitat This sticks out to me. The default language server port in Godot 3.x is `6008`, and in Godot 4 it's `6005`. Is `6007` a typo here, or have you intentionally modified your Godot editor settings to change the port to `6007`?

OsamaHeweitat:
> Very interesting. I did a quick test and I can't replicate this on `1.3.0`, `1.3.1`, or `master` connecting to Godot `3.5` or `3.5.1` on Windows 10.
> 
> > popup text:  "Connecting to the GDScript language server at 127.0.0.1:6007"
> 
> @OsamaHeweitat This sticks out to me. The default language server port in Godot 3.x is `6008`, and in Godot 4 it's `6005`. Is `6007` a typo here, or have you intentionally modified your Godot editor settings to change the port to `6007`?

I changed the port at the time to 6007 in hopes of the it working. The issue occurs with both 6008 and 6007 as the port. I tried adjusting a multitude of settings and nothing worked.

[...]

PR #439, opened 2022-12-12
Clean up plugin initialization process

DaelonSuzuka:
This should be a fix for #435.

Apparently, the plugin would fail to initialize if you were starting VSCode without a workspace. I exclusively open folders as workspaces, so I never thought to test this scenario.

The failure is because the plugin was trying to find `project.godot` when it activated, and several things assumed that would never fail. I've completely removed this search from the plugin activation and every place that needs the project file/directory now just looks it up when it needs it.

This also means I can remove a dependency (klaw), which is always good.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,6 @@
 			"dependencies": {
 				"await-notify": "^1.0.1",
 				"global": "^4.4.0",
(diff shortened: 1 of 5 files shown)


Calinou (approved):
Thanks!

PR #451, opened 2023-02-08
Add Godot 4 support to the Scene preview panel

DaelonSuzuka:
While working on a Godot 4 project, I noticed that the scene preview panel wasn't working correctly. It failed to parse the scene file, looking something like this:

Code_cokhC6WaeF

The problem was a number of regexes that were too specific, matching against the Godot 3 scene file but not the minor changes in Godot 4. Once parsed, the new classes weren't resolving to the set of available icons. 

After fixing the regexes, cleaning up some other issues, and integrating the icon set from Godot's master branch, now it looks like this:

Code_MvzeVdKkBo


DaelonSuzuka:
Also, I sincerely apologize for anybody attempting to view the 1800 file changeset. Github was barely responsive enough to create the PR in the first place.

Calinou:
The file naming for SVG files looks weird. For instance, it should be `AnimatedSprite2D.svg` instead of `Animatedsprite2D.Svg`. Node names should match the ones used in `editor/icons`, and file extensions should be lowercase.

DaelonSuzuka:
> The file naming for SVG files looks weird.

Excellent catch, thank you. I accidentally converted the entire filename instead of just the basename.

> Did you use a script for this?

Sorry,  I _intended_ to write about this in the PR body but GitHub was lagging so badly I could barely paste in the screenshots.

Yeah, I wrote a python script that converts the filenames of the existing icons, copies the godot 4 icons to the light and dark folders, and changes the colors to match what I think is the existing pattern for each theme.

I think I should rewrite that script in js, so it can be called with `npm run`. Perhaps a new `tools` directory? Then `npm run regenerate_icons` can:
- get the icons from both 3.x and 4.x
- (?) filter out the non-class icons
- fix the 3.x names
- copy them to the theme directories
- remap the colors

I believe @Razoric480 committed the existing set of icons, so I reached out to him to see if I missed anything about the conversion process.

Razoric480:
@DaelonSuzuka  @Calinou 
Back in the godotengine/godot-design repo, icons used to be stored in there back in the day, and that included an optimize.py script that included the light/dark color mapping and it would both clean the SVGs and recolor them.

https://github.com/godotengine/godot-design/blob/33740d504632e0630bc91f5319be102630791022/engine/icons/optimize.py

The re-coloring is now done in-engine.

DaelonSuzuka:
Alright, the new icons should all have fixed names, and I wrote a typescript tool that can generate the icons. It's registered in `package.json`, and can be called with `npm run generate-icons <path/to/godot/repo>`.

I tried to have the script download the icons from github but that was too hard. Instead, the script requires you to already have a clone of the engine repo on your system, and pass the path of that repo to the script when you run it.

It scrapes the icons from the `3.x` and `master` branches, and scrapes the contents of [godotengine/godot/blob/master/scene/register_scene_types.cpp#L315](https://github.com/godotengine/godot/blob/master/scene/register_scene_types.cpp#L315) to filter out the icons that aren't for actual classes. This cuts the icon count from 2028 to 542.

Calinou:
Is this ready to merge now? Also, does this keep compatibility with Godot 3.x?

DaelonSuzuka:
@Calinou Fully 3.x compatible and ready to merge.

Calinou (approved):
Thanks!

PR #457, opened 2023-03-04
Fixes for godot 4 scene preview

QuinsZouls:
This a set of fixes for the scene tree in godot 4 ref: #451 .
### What's include?
- Update typescript version
- Improve tsconfig for a better development experience (customs paths, allow esModule compatibility, allow SyntheticDefaultImports, an more!)
- Improve `generate-icon` script with the new tsrule and add CSG Icons generation
- Fix issue with the regex of `scene_preview_provider` that cause some regressions with godot 3 and 4
- Add CSG Icons
### Before patch
Captura de pantalla 2023-03-04 172904
### After patch
Captura de pantalla 2023-03-04 174111


ddelvall:
Any news about the update of the plugin?

DaelonSuzuka:
If there's feedback that's relevant to an open PR, the correct place for that is in the comments of that PR. It's not appropriate to make new PRs that completely contain existing open PRs.

One big reason is that involving the original author of a given PR can help avoid obvious errors, like:
- adding a second loop to check `modules/csg/icons` but not noticing that 5 other modules also have icons
- changing the behavior of the final "writing icons" section to no longer overwrite existing icon files, when the point is to make sure this extension's icons match the latest versions from the editor
- adding unrelated configuration changes that, as previously mentioned, belong in their own PR
- pointless changes like forEach -> for loops that have no impact to you since you aren't going be maintaining this code in the long run

It's unfortunate that this PR was submitted this way, because there were a number of real mistakes in my original code and you fixed several of them. If this feedback had been in the correct place, I'm sure we could have had a productive discussion, I could learn a lesson about proofreading, and both of us could have avoiding wasting our time with this whole episode.

PR #459, opened 2023-03-14
Fix errors left after cleaning up plugin initialization process

DaelonSuzuka:
#439 apparently had a couple accidents, a leftover log statement and a reference to a removed class variable, `workspace_dir`.

The reference to `workspace_dir` affects the code path for launching the Godot editor from VSCode, and that feature still works even when I remove the entire line. I didn't write this feature, and I don't see a purpose for this line in the first place.


Calinou (approved):
Thanks!

PR #452, opened 2023-02-10
Rewrite debugger for Godot 4 support + improved maintainability

DaelonSuzuka:
This is a continuation of https://github.com/godotengine/godot-vscode-plugin/pull/400 by @RedMser. I'm opening this as a draft PR so I have a place to keep some notes.

Current state:
> The main thing holding this back, besides some issues with the debugger, is compat with 3.x.
>
> Keeping compatibility with multiple Godot versions will probably double the size of this PR, due to having to create duplicates of every system (debugger, connection, etc.) which must be interchangeable. Not to speak of automatic version detection, which I'm not sure how it should be done just yet.

It may be possible to integrate the changes in #387 into this PR.

PR #461, opened 2023-03-29
Update: GDScript Debugger setup for vs code

codelinx:
- VS Code settings path changed in command palette.


DaelonSuzuka:
Thank you for identifying this problem. The debugger is currently undergoing a major rewrite in #452, and the README/docs will need to be overhauled once that's complete.

PR #469, opened 2023-04-27
Change default LSP Port for Godot 4's one

jpcerrone:
Changes the default LSP Port from 6008 (Used in Godot 3) for 6005 (Used in Godot 4). This change was made in https://github.com/godotengine/godot/commit/de7873c2d87c3ee8af8c27e35f49767fcc384e75 .

Also fixes a potential bug in the `check_client_status()` function. It was wrongly getting the `serverProperty `from the `serverPort `configuration and viceversa.

### DIFF
--- a/package.json
+++ b/package.json
@@ -181,7 +181,7 @@
 				},
 				"godotTools.lsp.serverPort": {
 					"type": "number",
-					"default": 6008,
+					"default": 6005,
 					"description": "The server port of the GDScript language server"
 				},
 				"godotTools.editorPath": {
--- a/src/godot-tools.ts
+++ b/src/godot-tools.ts
@@ -230,8 +230,8 @@ export class GodotTools {
 	}
 
 	private check_client_status() {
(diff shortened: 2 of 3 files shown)


xqrdot:
That would break the compatibility with Godot 3, which is an LTS version of Godot as of now. Ideally, automatic LSP port discovery should be a thing, with "6008" and "6005" being default options.

I'd suggest making a separate PR to fix the `check_client_status()` method: that fix can easily make its way to an upstream.

jpcerrone:
> Also fixes a potential bug in the `check_client_status()` function. It was wrongly getting the `serverProperty `from the `serverPort `configuration and viceversa.

Got it. I'm closing this PR and opening [](https://github.com/godotengine/godot-vscode-plugin/pull/472) to address the `check_client_status()` fix.

raylu:
perhaps we could check the version by running `${editor_path} --version`?

alternatively, a fallback port and have the extension try both ports in sequence?

PR #472, opened 2023-05-01
Fix check_client_status() getting port and host configurations mixed up

jpcerrone:
Fixes a potential bug in the check_client_status() function. It was wrongly getting the serverProperty from the serverPort configuration and viceversa.


DaelonSuzuka:
Thanks for identifying this issue. This was fixed as part of larger changes in #488.

Issue #473, opened 2023-05-02
Default LSP server port of plugin doesn't match with default port of Godot 4 


tymbaca:
### Godot version

4.0.2.stable.official

### VS Code version

1.77.3 (Universal)

### Godot Tools VS Code extension version

1.3.1

### System information

MacOS Monterey (12.6.3)

### Issue description

By default Godot 4 Language Server port is set to "6005", whereas Godot Tools default is set to "6008".

So can't connect to Godot 4's Language Server via VS Code without changing either Godot Tools settings or Godot Language Server settings.

### Steps to reproduce

- Install Godot Tools VS Code extension
- Install Godot 4 (make sure "Network/Language Server" settings are set to default
- Try to connect to LSP via VS Code

Calinou:
For reference, this is why the default port was changed:

> The LSP port changed in Godot 4, as the one used in Godot 3 conflicted with the normal debugger port when using multiple Godot instances.

https://github.com/godotengine/godot/issues/75683

idbrii:
FYI, these lines in godot-vscode-plugin need to be changed: https://github.com/godotengine/godot-vscode-plugin/blob/master/package.json#L182-L186. Found while I was fixing vim-lsp-settings, but I don't use vscode so I'm not making a PR.

Godot 3 users of this plugin can change to use port 6005 in editor -> editor settings -> network -> language server -> remote port.

Here's the port number in Godot4's code:
https://github.com/godotengine/godot/blob/master/modules/gdscript/language_server/gdscript_language_server.h#L47-L48

lostfictions:
Hi there, can this issue be pinned in the interim? The impression this gives to users at the moment is that this extension simply doesn't work. There's nothing in the README about this.

jst4rk:
I have the same port configured in both and does not work at all, I even change the porotol to ws but nothing happens
I'm using the latest version of VSCode and Godot 4.1

CuitGGyy:
I have the same situation on my macos ventura. Godot 4.1.1 and godot-tools 1.3.1, I have configured  port 6005 and vscode connected godot, but when launch debug vscode still report error "Failed to launch Godot instance: Error: Command failed: ... "

midblue:
Can confirm that the address is set to 127.0.0.1:6005 in both Godot and VS Code and VS Code hangs forever on "Initializing" with the message "Connecting to the GDScript language server at 127.0.0.1:6005". (Godot 4.1, OSX)

[...]

PR #479, opened 2023-06-16
add warning and instructions to README.md for godot 4

lostfictions:
This PR attempts to alleviate some of the confusion caused by https://github.com/godotengine/godot-vscode-plugin/issues/473.

It adds a warning at top of README.md about the manual configuration required to use this extension with Godot 4, as well as instructions in the VS Code configuration section about what port to use.

I've lightly reformated that section to better allow for multi-line setting descriptions and anchor linking (so that the warning at the top of the readme can directly link to the relevant section).

Additionally, I removed the `checkStatus` setting documentation, since as far as I can tell it no longer exists. There seem to be a number of new, undocumented extension settings, but maybe those could be added in a further PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,12 @@
 A complete set of tools to code games with
 [Godot Engine](http://www.godotengine.org/) in Visual Studio Code.
 
+> **Warning**
+> 
+> This plugin requires manual configuration to work with Godot 4!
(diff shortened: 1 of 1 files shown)


DaelonSuzuka:
Thanks for doing this. The Godot 4 update is taking longer than expected (sorry!), so this is a good intermediate measure.

Also congrats on your first contribution!

PR #484, opened 2023-07-30
Update server_controller.ts

Rakaneth:
Add protocol for --remote-debug parameter


DaelonSuzuka:
Thank you for identifying this problem. The debugger is currently undergoing a major rewrite in #452, and this fix is going to be included in that PR.

PR #491, opened 2023-09-15
Improve debugger setup instructions

moritz-t-w:
just a lil thing, I noticed that the command specified doesn't show up, but this works.


DaelonSuzuka:
Looks good to me, thanks for contributing!

PR #492, opened 2023-09-16
Default language server port doesn't match with Godot 4 (#473) fix + Godot 3 back-compatibility

runny-yolk:
I figured it would be good to update the default for Godot 4 as people seem to get caught up on it fairly frequently (#473), however there's still a lot of v3 users, so as a compromise I've added a fallback port, which it will use when it's used up half of the max connection attempts. I also doubled the number of connection attempts, so both ports get the 10 they were getting before.

By default, the main `serverPort` is the Godot 4 default, and the fallback is the Godot 3 default.

I tested this code of course, but I haven't developed much for VScode or Godot before, so please scrutinise my code as much as you like.

As an aside, I noticed that the warning message at the top of the README isn't on the VSCode marketplace? Seems like the version published there might be out of date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,9 +153,6 @@ When developing for the extension, you can open this project in Visual Studio Co
 ### Why does it fail to connect to the language server?
 
 - Godot 3.2 or later is required.
-- For Godot 4, the [`gdscript_lsp_server_port` setting](#gdscript_lsp_server_port)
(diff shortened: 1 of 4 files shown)


runny-yolk:
@DaelonSuzuka Ah, no worries, I'm just happy to see the issue fixed.

PR #506, opened 2023-10-12
Multiple highlighting improvements

DaelonSuzuka:
- binary int notation
- `_` spacers in numeric literals
- fixed several instances of incorrect case insensitivity
- @"a", &"b", ^"c", and r"d" style string literals

### DIFF
--- a/syntaxes/GDScript.tmLanguage.json
+++ b/syntaxes/GDScript.tmLanguage.json
@@ -132,7 +132,23 @@
                     "end": "'''"
                 },
                 {
-                    "begin": "\"",
+                    "begin": "(r)?\"\"\"",
+                    "end": "\"\"\"",
+                    "patterns": [
+                        {
+                            "name": "constant.character.escape.untitled",
+                            "match": "\\\\."
+                        }
+                    ],
+                    "beginCaptures": {
+                        "1": {
+                            "name": "constant.character.escape.untitled"
+                        }
+                    },
+                    "name": "string.quoted.triple.gdscript"
+                },
+                {
+                    "begin": "(r)?\"",
                     "end": "\"",
                     "patterns": [
                         {
@@ -140,10 +156,15 @@
                             "match": "\\\\."
                         }
                     ],
+                    "beginCaptures": {
+                        "1": {
+                            "name": "constant.character.escape.untitled"
+                        }
+                    },
                     "name": "string.quoted.double.gdscript"
                 },
                 {
-                    "begin": "'",
+                    "begin": "(r)?'",
                     "end": "'",
                     "patterns": [
                         {
@@ -151,6 +172,11 @@
                             "match": "\\\\."
                         }
                     ],
+                    "beginCaptures": {
+                        "1": {
+                            "name": "constant.character.escape.untitled"
+                        }
+                    },
                     "name": "string.quoted.single.gdscript"
                 },
                 {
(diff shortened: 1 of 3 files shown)


AlfishSoftware:
Will this fix the `lambda_declaration` rule?
Because I was getting `support.function.any-method.gdscript` (method color) for `func` in `func(): ...` instead of `storage.type.function.gdscript` (keyword color).

AlfishSoftware:
Is `r"..."` actually valid syntax in GDScript?
I tested in Godot 3 and 4 script editors and neither is accepting it.
It's also not listed [here](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_basics.html#literals).

syntaxes/GDScript.tmLanguage.json
```diff
@@ -502,6 +535,30 @@
...
+            },
+            "patterns": [
+                {
+                    "match": "[^%^\"^']*",
+                    "name": "constant.character.escape"
+                }
+            ]
+        },
         "builtin_get_node_shorthand_quoted": {
             "begin": "(\\$)([\"'])",
```
AlfishSoftware, line 563: Godot 4 also accepts `%"..."` and `%'...'` quoted syntax.

DaelonSuzuka:
> Is r"..." actually valid syntax in GDScript?

https://github.com/godotengine/godot/pull/74995

syntaxes/GDScript.tmLanguage.json
```diff
@@ -309,37 +335,41 @@
...
                     "name": "constant.numeric.float.gdscript"
                 },
                 {
-                    "match": "\\b(?i:(\\d+e[\\-\\+]?\\d+))\\b",
+                    "match": "\\b(?:([0-9_]+e[\\-\\+]?\\[0-9_]))\\b",
                     "name": "constant.numeric.float.gdscript"
                 },
                 {
-                    "match": "\\b\\d+\\b",
+                    "match": "\\b[0-9_]+\\b",
```
AlfishSoftware, line 372: It's not right to allow `_` to intermix with numbers unconditionally. Cases like `_0` are valid variable identifiers, and must not be colored like number literals. This regex need to be carefully thought out and tested, because adding the `_` separator feature is not that simple. You need to ensure:
- `_` cannot appear at the beginning, or right after `e` in floats or `x` in hex literals
- `_` should not appear at the end, but it seems Godot allows it; better confirm this properly to know how to go about this
- consecutive `__` is not allowed

AlfishSoftware:
> > Is r"..." actually valid syntax in GDScript?
> 
> [godotengine/godot#74995](https://github.com/godotengine/godot/pull/74995)

Ah I see, so it's a new feature.
> "escaping" in r-strings exists, but the sequences are escaped into themselves

I've never used this r-string before, but if I understood correctly, it seems some care needs to be taken about the internal patterns, because I believe nothing inside should be colored like an escape character. It would probably need a separate rule(s).
For example, in `r"\""` both the `\` and the `"` are part of the string, even though `"` won't close it when preceded by `\`.

So it should look like the image [in this comment](https://github.com/godotengine/godot/pull/74995#issuecomment-1473293698) in those tests:
r-string coloring

syntaxes/GDScript.tmLanguage.json
AlfishSoftware: On this particular line it would probably be enough to use `"\\b\\d[0-9_]*\\b"` if you won't forbid trailing `_`. On float and hex literals it'd probably be similar.

DaelonSuzuka: I'm not interested in completely validating numeric literals with the grammar. Invalid cases will be syntax errors and the LSP should mark them as such.

AlfishSoftware: That's fair for invalid cases that aren't valid for other rules with less priority. But I believe valid variable names like `_0`, `_0e0`, `_0x0` will be taken over by this rule and be incorrectly colored as numbers. At the bare minimum, these regexes gotta exclude the initial `_` by using something like `\d[0-9_]*` instead of `[0-9_]+`.

But if you don't mind me completely validating it, the ideal would be:
int: `"\\b\\d+(?:_\\d+)*\\b"`
hex: `"\\b0x\\h+(?:_\\h+)*\\b"`
bin: `"\\b0b[01]+(?:_[01]+)*\\b"`
float1: `"\\b\\d+(?:_\\d+)*\\.(?:\\d+(?:_\\d+)*)?(?:[eE][-+]?\\d+(?:_\\d+)*)?\\b"`
float2: `"\\.\\d+(?:_\\d+)*(?:[eE][-+]?\\d+(?:_\\d+)*)?\\b"`
float3: `"\\b\\d+(?:_\\d+)*[eE][-+]?\\d+(?:_\\d+)*\\b"`

This should also fix:
- `[eE]` exponent which can be uppercase too
- a typo where the `\\` part of `\\d+` on line 338->368 wasn't removed
- float2 starts with `.` (non-word character) so it must not be preceded by a `\b` word boundary

Sorry for writing this in a comment instead of making a PR or something.

DaelonSuzuka: > ... other rules with less priority. But I believe valid variable names like _0, _0e0, _0x0 will be taken over by this rule ...

Ah, I'm with you now; I misunderstood the problem. 

I'll add some new test cases and then pop in your rules. Thanks for writing them up!

DaelonSuzuka:
> r-string stuff

Makes sense. I was just making sure they were highlighted at all.

> screenshot

Wait a second, are triple single quotes valid in Godot 4? IIRC they're illegal in Godot 3 and I have them marked as such.

syntaxes/GDScript.tmLanguage.json
DaelonSuzuka: Good catch.

DaelonSuzuka:
> Will this fix the `lambda_declaration` rule? Because I was getting `support.function.any-method.gdscript` (method color) for `func` in `func(): ...` instead of `storage.type.function.gdscript` (keyword color).

This is what I'm currently seeing:

image

AlfishSoftware:
How strange ... this is what I'm getting (both when Godot is connected or disconnected).
lambda_declaration rule issue screenshot
I'm on `geequlim.godot-tools v1.3.1` on VSCodium v1.84.2

DaelonSuzuka:
I think I took those screenshots with this PR, not `1.3.1`. Would you might checking with `master` to make sure that this is working correctly for you?

> (both when Godot is connected or disconnected)

Why would this matter?

DaelonSuzuka:
It could, if the Godot LSP supported semantic tokenization, but it doesn't.

Calinou:
> Wait a second, are triple single quotes valid in Godot 4? IIRC they're illegal in Godot 3 and I have them marked as such.

Triple single quotes have always been allowed, even in Godot 1.x.

DaelonSuzuka:
> > Wait a second, are triple single quotes valid in Godot 4? IIRC they're illegal in Godot 3 and I have them marked as such.
> 
> Triple single quotes for multiline strings have always been allowed, even in Godot 1.0.

For some reason, 2 years ago I thought it wasn't. Maybe I wrote a python-style docstring, got an error, and assumed, who knows?

Either way I've fixed it now.

PR #515, opened 2023-10-16
Syntax highlighting changes

ryanabx:
* Adds the upcoming `@abstract` annotation https://github.com/godotengine/godot/pull/82987 (4.3) to the list of annotations in the gdscript textmate grammar. ~~will move to a new PR~~ nvm
* Complete syntax highlighting for properties. Class properties used to be unhighlighted (the only properties highlighted were `.` properties) and are now highlighted properly as `variable.other.property.gdscript`
* Treat `void` as a `keyword` instead of a `class`. See https://github.com/godotengine/godot/blob/a574c0296b38d5f786f249b12e6251e562c528cc/modules/gdscript/gdscript_tokenizer.h#L130C4-L130C8 (gdscript considers `void` to be a keyword, similar to C-style languages)

NOTE: This is my first time working with regular expressions, so please let me know if I've made a mistake with my added regular expressions!

Before:
image

After:
image

### DIFF
--- a/syntaxes/GDScript.tmLanguage.json
+++ b/syntaxes/GDScript.tmLanguage.json
@@ -94,6 +94,9 @@
             "include": "#any_method"
         },
         {
+            "include": "#any_variable"
+        },
+        {
             "include": "#any_property"
         },
         {
@@ -375,7 +378,7 @@
             ]
         },
         "variable_definition": {
-            "begin": "\\b(?:(var)|(const))\\s+",
+            "begin": "\\b(?:(var)|(const))\\s+([a-zA-Z_]\\w*)\\s*",
             "end": "$|;",
             "beginCaptures": {
                 "1": {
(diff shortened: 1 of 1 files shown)


atirut-w:
The new annotation should probably be in a separate PR that depends on https://github.com/godotengine/godot/pull/82987 being merged. The other two seems great for consistency.

DaelonSuzuka:
> The new annotation should probably be in a separate PR that depends on https://github.com/godotengine/godot/pull/82987 being merged.

No, there's no reason to push that out to another PR. Adding and removing annotations is a trivial change, and the linked engine PR seems quite likely to get merged.

The grammar in this extension is explicitly attempting to be a superset of all versions of GDScript, so adding support for future features is just as valid as maintaining support for Godot 3 features or even features that were deprecated in 3.2 and removed in 3.5.

DaelonSuzuka:
Sorry for the delay. Everything looks fine to me.

PR #526, opened 2023-11-12
Adjust Godot version detection regex

mlsvrts:
The previous version detection would fail to detect some versions of godot (e.g 4.1.1.stable.mono.official.bd6af8e0e), resulting in an unclear error message. This commit makes a minor change to slightly improve the robustness of the version detection regex.

With this change I can run and debug my project with godot 4.1.1, and breakpoints even work 😄

### DIFF
--- a/src/debugger/godot3/server_controller.ts
+++ b/src/debugger/godot3/server_controller.ts
@@ -109,7 +109,7 @@ export class ServerController {
 		try {
 			log.info(`Verifying version of '${godotPath}'`);
 			const output = execSync(`${godotPath} --version`).toString().trim();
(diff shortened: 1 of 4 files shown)


DaelonSuzuka:
Thanks for catching that, and congrats on your first contribution!

mlsvrts:
> Thanks for catching that, and congrats on your first contribution!

No problem! I was very happy to find that, besides this small issue, the debugger upgrade has made Godot 4 debugging integration very painless; thanks for your work on it!

DaelonSuzuka:
> the debugger upgrade has made Godot 4 debugging integration very painless; thanks for your work on it!

That's great to hear! There's more work to do, but it should way more stable and reliable than before.

PR #529, opened 2023-11-14
Various QoL Improvements

DaelonSuzuka:
* Updated Godot icons (and removed old ones)
* Improve "Debug Pinned Scene" command by making the pinned scene persist between VSCode sessions
* Fix scene file highlighting not working 
* Fix ScenePreview not working in Godot 3
* Add buttons for relevant actions to ScenePreview items
* Add internal document links for SubResource() and ExtResource() statements in scene files
* Add hover for SubResource() and ExtResource() statements in scene files
* Improve ability to right click -> open docs for methods of builtin types (doesn't always work)
* Add a file decorator to show the pinned debug file in the filesystem view/editor tabs
* Add item decorators to the Scene Preview to show Node attributes more clearly
* Updated readme
* Overhauled documentation viewer
* Added GDScript formatter
* Add (disabled) experimental providers for custom completions, semantic tokens, and tasks


sandygk:
I am seeing an issue on this branch that I am not seeing on master, I am getting no icons on the godot scene preview

here is how it looks on this branch
image


and this is master for reference

image


I even pull master from this branch, but still got the same, maybe it's something I am doing wrong, or it may be a known issue

DaelonSuzuka:
@sandygk Honestly that's pretty weird. This _has_ happened to me but I seem to recall it was a local problem. Try stopping the "Watch" process in the terminal, deleting the `out/` folder in the root directory, and restarting the extension host.

sandygk:
> @sandygk Honestly that's pretty weird. This _has_ happened to me but I seem to recall it was a local problem. Try stopping the "Watch" process in the terminal, deleting the `out/` folder in the root directory, and restarting the extension host.

I tried that, but it didn't fix it, I removed the `out/` dir and ran `npm run package` and got the same result, when I change to master and run `npm run package` and install that version it works fine. 

It's weird, because I noticed a similar comment was made for a different branch [here](https://github.com/godotengine/godot-vscode-plugin/pull/452#issuecomment-1807169710), but I ran it after it was merged into master, and it works fine for me on master

sandygk:
Ok so I tried to debug this and I noticed something that may be of help. If I run the extension in debug mode, it works fine
image

but if I run `npm run package` and install it, then I don't have the icons

DaelonSuzuka:
That makes a lot more sense! I changed some typescript settings recently, I bet that messed up the bundler. Should be an easy fix.

sandygk:
Another thing I noticed in this branch: I have VSCode set up to format on save, with the main branch it doesn't do anything, but with this branch it tries to fix the formatting, but it breaks the indentantion in a bunch of places:

image

DaelonSuzuka:
Yeah there's a new formatter and it's somewhat experimental. Mixed tabs and spaces and correctly preserving indentation is something it doesn't handle well yet. As a stopgap I think I can just make it not touch leading whitespace at all, until I figure that out.

sandygk:
I'm using this branch for development. I'll post if I find any other issues with the formatter or something else.

sandygk:
> Mixed tabs and spaces and correctly preserving indentation is something it doesn't handle well yet. As a stopgap I think I can just make it not touch leading whitespace at all, until I figure that out.

I think it may be better to leave that to VSCode, it already has features to handle tabs and spaces:

image

DaelonSuzuka:
@sandygk The scene preview icons should be fixed now. The "Active Scene Tree" view in the debugger also had broken icons, which should be fixed too.

I'm not exactly sure what the problem was because I reverted my tsconfig changes and it didn't fix it. Best guess is that I've reorganized modules to use `index.ts` for cleaner imports, and that's causing the bundler to be invoked differently that just folders + files.

src/lsp/DefinitionProvider.ts
```diff
@@ -0,0 +1,39 @@
...
+export class DefinitionProvider implements vscode.DefinitionProvider {
+
+	public data: Map<string, string> = new Map();
+
+	constructor(private client) {
+		vscode.languages.registerDefinitionProvider({ scheme: "file", language: "gdscript" }, this);
+	}
+
+	async provideDefinition(document: vscode.TextDocument, position: vscode.Position, token: vscode.CancellationToken): Promise<vscode.Definition> {
+		const key = `${document.uri},${position.line},${position.character}`;
```
levidavidmurray, line 15: This is where the intermittent issue come from when go to definition doesn't open the docs.

Taking inspiration from my original PR, you could instead get the symbol via a mock `textDocument/hover` message. I'd personally extract the symbol parsing logic from `handle_hover_response` into its own method and then you can return the symbol from an async `get_symbol_at_position` method called from `DefinitionProvider::provideDefinition`

```typescript
// GDScriptLanguageClient.ts
public async get_symbol_at_position(uri: vscode.Uri, position: vscode.Position): string {
    const params = {
	  textDocument: { uri: uri.toString() },
	  position: { line: position.line, character: position.character },
    };
    const response = await this.sendRequest("textDocument/hover", params);
    return this.parse_hover_response(response)
}
```

```typescript
// DefinitionProvider.ts
export class DefinitionProvider implements vscode.DefinitionProvider {
    
    constructor(private client: GDScriptLanguageClient) {}
    
    async provideDefinition(document: vscode.TextDocument, position: vscode.Position, token: vscode.CancellationToken): Promise<vscode.Definition> {
        const symbol = await this.client.get_symbol_at_position(document.uri, position);
        if (!symbol) {
            return null;
        }
        
        const parts = symbol.split(".");
        const uri = vscode.Uri.from({
            scheme: "gddoc",
            path: parts[0] + ".gddoc",
            fragment: parts[1],
        });
        
        return new vscode.Location(uri, new vscode.Position(0, 0));
    }
}
```

```typescript
// ClientConnectionManager.ts
public client: GDScriptLanguageClient = null;
public definitionProvider: DefinitionProvider = null;

constructor(private context: vscode.ExtensionContext) {
    ...
    this.client = new GDScriptLanguageClient(context);
    this.definitionProvider = new DefinitionProvider(this.client);
    vscode.languages.registerDefinitionProvider({ scheme: "file", language: "gdscript" }, this.definitionProvider);
    ...
}
```

It seems to work pretty flawlessly from the little testing I've done so far.

sandygk:
@DaelonSuzuka I found another issue with the formatter, it seems to have issues formatting lambda functions, in general it, tend to remove more spaces than usual, and in some cases it just breaks the code, here is an example:

it turns this:

image

into this:
image

Also the return type for lambdas is not colored, but that is happening in master as well

DaelonSuzuka:
@sandygk Try it now.

image

DaelonSuzuka:
Once more, with feeling.

sandygk:
image
working like a charm!

DaelonSuzuka:
# New things added today

I'm partway through fixing broken code boxes in hovers and docs.

Inlay hints to help reading scene files:

Code_YYyYfShnDu

`Go to Definition` now works in scene files:

Code_lf83ELz3qV

src/formatter/tokenizer.ts
```diff
@@ -0,0 +1,301 @@
...
+		if (prev === "node") return " ";
+		return "";
+	}
+
+	let lastTokenType: string;
+	let token: string;
+	const tokens: string[] = [readWhitespace()];
+	while (pos < input.length) {
+		const char = input.charAt(pos);
+		const nextChar = input.charAt(pos + 1);
```
sandygk, line 204: Even though we have the variable `nextChar` we are using `input.charAt(pos + 1)` in a few places, should we replace it with the variable instead for consistency and readability sake?

image

DaelonSuzuka: Yeah that's a good catch. 

This entire formatter is basically a placeholder until I have time to finish the real one.

sandygk:
I noticed another minor issue, I think it was introduced recently, it's adding one space before not always, 

image

DaelonSuzuka:
This looks like it'll be a harder one to untangle without a deeper rewrite than I'd like. In the meantime you could use `!` instead of `not`?

sandygk:
That's ok, do you want me to create a bug for it once the PR is merged?

DaelonSuzuka:
No, I'm still gonna fix it, it just won't be within the hour like I usually manage.

DaelonSuzuka:
@sandygk I completely rewrote the formatter. This new version uses the TextMate engine for parsing, and it's able to use the GDScript grammar we already use for syntax highlighting.

As unexpected but pleasant surprise, I immediately found several bugs in the GDScript syntax highlighting.

sandygk:
@DaelonSuzuka, nice, I found a couple of issues with the new version of the formatter, here are undesirable changes introduced by the new formatter on my current project, some even break the code.

image

image
image
image
image
image
image
image
image
image
image
image
image
image
image



It does fix the issue with the space before `not` though

DaelonSuzuka:
Fixed:

- await
- $Node and %Node things
- -number things (this is surprisingly difficult, probably still not correct)
- enum {
- if ( and and (
- thing(); thing();

I don't think `Utils.chance(1, /3)` is valid syntax

Having less whitespace inside function parens is intentional.

sandygk:
>Utils.chance(1, /3)

It's a dot after the 1, not a comma

sandygk:
Here are the issues I am getting now on my project:

image
image
image
image
image
image
image
image
image
image
image
image
image
image

sandygk:
when you do func() it colors it well, but if you do `func ()` the color breaks

sandygk:
It would be good if you could get your hands on large godot 3 and 4 repos to run the formatter on, that way you can check the diff that it gives you as a form of testing. If you want I can send you mine, but it's not that large to be honest.

sandygk:
We can build a gdscript with all the syntax that we can think of, and make sure we format it correctly. We can run the formatter on it and use it as a unit test, that way we can quickly check that one fix doesn't break another.

DaelonSuzuka:
I've rewritten the documentation handler. The previous "Open Godot Documentation" context menu command has been removed, everything is handled via go-to-def and `.gddoc` custom file handling. 

It works for me when I move the cursor with arrow keys and hit F12, so please let me know if there's any issues with your vim movements, @levidavidmurray 

As a cool bonus, we get this behavior: 

Code_tqplQfnTsy

Because the documentation is now a virtual custom file, reloading the editor will cause vscode to automatically reload the docs, too (assuming that it connects to the LSP in a timely manner, of course).

PR #541, opened 2023-11-29
Open documentation from "go to definition" on native symbols

levidavidmurray:
This has been bothering me for a while, so figured I'd do something about it.

https://github.com/godotengine/godot-vscode-plugin/assets/46084870/2f087dd0-5ef7-41d8-b6eb-525ce2f4ea69

Added the following settings to control this behaviour:

- `godotTools.documentation.method` (`web`, `webExternal` or `local`)
- `godotTools.documentation.webUrl` (_default: https://docs.godotengine.org/en/stable_)

The GDScript LSP can sometimes struggle to resolve a symbol and returns multiple types

image

You _can_ cast the value to get around this, but I opted to add a QuickPick menu to allow you to select from the given options

https://github.com/godotengine/godot-vscode-plugin/assets/46084870/dc21a637-445a-4ab8-9f94-c1d9477f3328

- [ ] Test with Godot 3.x LSP
- [ ] Fix hovering a symbol with CTRL held (no clicking) triggering documentation open

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ experience as comfortable as possible:
 - In-editor Scene Preview
 - Display script warnings and errors
 - Ctrl + click on a variable or method call to jump to its definition
-- Full documentation of the Godot Engine's API supported (select *Godot Tools: List native classes of Godot* in the Command Palette)
+- Open Godot documentation when Ctrl + clicking native classes, methods, and properties
 - Run a Godot project from VS Code
 - Debug your GDScript-based Godot project from VS Code with breakpoints, step-in/out/over, variable watch, call stack, and active scene tree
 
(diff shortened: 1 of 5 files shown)


levidavidmurray:
> The "bad" news is that this feature has already been implemented in https://github.com/godotengine/godot-vscode-plugin/pull/529:

Haha, all good! This was mostly a project taken on to procrastinate working on my game 🤡 Definitely should've taken a closer look at any existing PRs.

I use VSCodeVim and didn't realize the problems my solution has with CTRL+clicking (I just use `g+d`). Hooking into the `textDocument/definition` message isn't a viable solution because there could be plenty of cases where that message occurs without going to definition (such as holding CTRL and hovering over a symbol).

I see you did things the correct way and created an actual `vscode.DefinitionProvider`! I did have some problems getting some things working properly, but I'll close this out and move the discussion to your PR.

PR #544, opened 2023-12-07
Fix variables with leading underscores

DenisBelmondo:
The integer literal regex was breaking variables with leading underscores. This PR should fix that.


Calinou (approved):

DaelonSuzuka:
This was already fixed in #545, plus the same issue for floats. 

I'm sorry I missed this PR, the last couple of weeks have been extra busy.

PR #557, opened 2023-12-22
Fix BBCode [br] not rendering in hover

DeszkaCodes:
Fixes the issue of [br] bbcode's weren't rendering new lines.

### DIFF
--- a/src/lsp/GDScriptLanguageClient.ts
+++ b/src/lsp/GDScriptLanguageClient.ts
@@ -154,6 +154,7 @@ export default class GDScriptLanguageClient extends LanguageClient {
 					value = value.replace(/\n[#]+/g, "\n");
 
 					// fix bbcode code boxes
+					value = value.replaceAll("`br`", "\n\n");
 					value = value.replace("`codeblocks`", "");
 					value = value.replace("`/codeblocks`", "");
 					value = value.replace("`gdscript`", "\nGDScript:\n```gdscript");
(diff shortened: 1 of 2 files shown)


DaelonSuzuka:
Can you share some examples of where that happens so I can validate the behavior + the fix?

DeszkaCodes:
> Can you share some examples of where that happens so I can validate the behavior + the fix?

Of course.

This happened every time I added "##" documentations to functions and hovered over them. `[br]` tags rendered like they were inside a `[code]` tag.

For example:
```
## A test function. [br]
## Testing.
func test():
    return true
```

Rendered hover documentation like:
"A test function. `br` Testing."

While the expected behaviour is:
"""
A test function.
Testing.
"""

DaelonSuzuka:
Oh, I get it. I didn't notice at first that *you* were adding the `[br]` tag, I was trying to think of why the language server would send bbcode line breaks and I was very confused.

Are you doing that to work around the fact that Godot 4 seems to not separate multiple lines of doc comments?

image

@Calinou @ryanabx Is this a bug in the G4 language server? G3 returns doc comments without concatenating them:

image

DaelonSuzuka (approved):

DeszkaCodes:
> Are you doing that to work around the fact that Godot 4 seems to not separate multiple lines of doc comments?

Yep, that's what I meant to fix somehow.

DaelonSuzuka:
@DeszkaCodes Thanks for your contribution, and congrats on getting your first PR merged here!

PR #571, opened 2024-01-19
fix Godot4.x debug console print multiple logs into one line.

kaluluosi:
Before fix
Snipaste_2024-01-19_20-18-53

After fix
Snipaste_2024-01-19_20-11-33

The server_controller of Godot 3 has no this bug.  So only fix for Godot 4.


DaelonSuzuka:
Looks good, thanks!

Issue #588, opened 2024-02-10
Add support for type inference inlay hints


LeoDog896:
### Godot version

4.2

### VS Code version

1.86.0

### Godot Tools VS Code extension version

2.0.0; bf1d739d073c93f62abf280a7ae19d5c3054c935

### System information

Windows 10

### Problem statement

You can't see the type of inferred assignments without having to hover over the variable or the function that creates it

### Proposed solution

Use VSCode's native [inlay hints](https://code.visualstudio.com/docs/typescript/typescript-editing#_inlay-hints) feature.

### Extra notes:
- I don't mind implementing this
- This extension has never implemented inlay hints, so it doesn't seem as if the platform information above matters.

Calinou:
Sounds good to me. I suggest coordinating with @DaelonSuzuka on this, but it may be better to focus on finalizing what's already being developed so we can release the long overdue 2.0.0.

DaelonSuzuka:
Not a bad idea. I investigated inlay hints for types before, but the type information that's available from the LSP is inconsistent. I wasn't sure how good the experience would be if only half or less of the variables had hints resolve properly.

> I don't mind implementing this

If you *want* to do it you're more than welcome. I'm already familiar with the systems required so it would probably take me an hour or two to add this.

Implementation notes off the top of my head:
- This feature would need to work with Godot 3 and 4, and the relevant LSP messages are formatted differently between the two.
- Some users HATE inlay hints, so there needs to be a setting that disables it
- `src/providers/inlay_hints.ts`

> This extension has never implemented inlay hints, so it doesn't seem as if the platform information above matters.

Open a scene file lol:

![image](https://github.com/godotengine/godot-vscode-plugin/assets/18042232/dfadff3d-fc80-47bd-adcb-5d0a05cb3f20)

LeoDog896:
> Open a scene file lol:

whoops! meant in GDScript;

DaelonSuzuka:
> whoops! meant in GDScript;

I assume most users have never even opened a scene file in VSCode, so I have to take every opportunity to shill all the work I put into those features!

LeoDog896:
> > whoops! meant in GDScript;
> 
> I assume most users have never even opened a scene file in VSCode, so I have to take every opportunity to shill all the work I put into those features!

[...]

PR #589, opened 2024-02-11
Add basic inlay hint support for GDScript

LeoDog896:
Adds #588, creating inlay hints in GDScript files

image

Currently works with Godot 3.x and 4.x (tested on the latest stable of both; 4.2 and 3.5.3)

### DIFF
--- a/package.json
+++ b/package.json
@@ -303,6 +303,16 @@
 					],
 					"default": "sameFolder",
 					"description": "Controls where the Scene Preview will search for related scenes when viewing a script file."
+				},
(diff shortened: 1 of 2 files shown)


DaelonSuzuka:
The basic inlay hints mechanism works, but there are some pathological cases where types are resolved incorrectly. Further incremental improvements can be done in a future PR.

I've set the GDScript inlay hints to be disabled by default and noted it as an experimental feature in the setting description.

Thank you @LeoDog896, and congratulations on your first contribution here. I look forward to working with you again!

PR #628, opened 2024-03-20
Fix poor documentation formatting of class titles and inheritance chain

jogly:
It seems the extended inheritance chain previously being parsed is no longer available, which causes the "Inherits: " information to be empty. Inspecting the symbol information available, it seems there is no way to recover the inheritance chain without additional queries or modifying LSP response, so a quick fix to display the correct information that we do have is an incremental improvement.

**Before**
image

**After**
image


DaelonSuzuka:
@jogly Thanks for catching this and for the quick fix.

PR #629, opened 2024-03-20
Improve codeblock formatting in documentation

jogly:
Improves codeblock highlighting by:

 -  Adding context to the codeblock detailing which language the following code is.
 -  Setting the `pre` block's background to match the default background for `code` blocks to give a consistent experience.
 -  Setting a border radius for the `pre` block double that of the `code` block's default (4px, which is hard-coded).

Works on class documentation blocks and `li` code blocks for properties/members/methods.

**Before**

image
image

**After**

image
image


DaelonSuzuka:
That's awesome, thank you.

PR #635, opened 2024-03-31
Fix project not found when `project.godot` file is excluded

InfiniteXyy:
Fix a problem that when `project.godot` is excluded in VSCode, `get_project_version` function will always return `undefined`

**How to reproduce:**

1. Add `project.godot` into vscode settings
image
2. Run **open workspace with Godot Editor**

There will be an error popup that the current workspace is not a godot project

**There is a simple fix:**

The `findFiles` api support taking a `null` value in the second parameter, then it will not take any exclusion from user config.

https://code.visualstudio.com/api/references/vscode-api

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -17,7 +17,7 @@ export async function find_file(file: string): Promise<vscode.Uri | null> {
 		return vscode.Uri.file(file);
 	} else {
 		const fileName = path.basename(file);
-		const results = await vscode.workspace.findFiles("**/" + fileName);
+		const results = await vscode.workspace.findFiles("**/" + fileName, null);
 		if (results.length == 1) {
 			return results[0];
 		}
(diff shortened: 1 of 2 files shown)


DaelonSuzuka:
Why on earth would you want to exclude the extremely important project.godot?

InfiniteXyy:
> Why on earth would you want to exclude the extremely important project.godot?

Hello, when I first tried godot, I didn't know what project.godot did. I thought that it was a program-generated file. 

Therefore, I excluded them and only keeps the gd files. 

I did something wrong, but the plugin keeps reporting meaningless errors, and I don't know what happens.

I think it should be a bug, even the user exclude the project file, the plugin should read the hidden files (maybe for some personal preference or just because unfamiliar with Godot)

atirut-w:
Wouldn't a better choice be disabling that popup with its explicitly excluded?

InfiniteXyy:
In TypeScript project, when I exclude the `tsconfig.json`, the ts plugin is still able to detect and use that config file.

I think here **files.exclude** only means not showing in the tree structure

DaelonSuzuka:
Thanks for submitting a fix for this, @InfiniteXyy.

PR #661, opened 2024-05-24
Improve dragging items from Scene Preview into source code

iMarv:
Commit message:
```
The right-click action, that drops the node-path into your clipboard already provides close to the same utility of the drag-n-drop feature - only missing a `$`.
To complement this and to add another QoL change, I refactored the drop-edit functionality to include a full `@onready` template. This will now create a variable name based on the node-name and provides the full convention, inspired by the same feature in the godot editor: https://github.com/godotengine/godot-proposals/issues/4482
```

The mentioned feature in the godot editor by default drops the $-import aswell, but allows dropping a full onready template when holding ctrl.
As far as I understand, VsCode does not allow fetching which keys the user may be pressing, so duplicating this feature 1:1 is not possible.

The extension already has amazing support for autocompleting $-imports in intellisense, so for me a big timesaver would be to be able to drag-n-drop the entire onready statement from the scene tree.

I have added this change as a reference implementation in this Pull Request. As far as I understand, there is no equivalent to this in C#, so I left the GetNode string there untouched.

Is this a feature to consider as the default? If not, would you be open to add it as a config option to the app?


DaelonSuzuka:
How about this:

Code_ltaV0QOWrx

iMarv:
🤯 great Idea
I also discovered that the Godot Editor supports this for uniquely named Nodes, which I would like to add in the next couple days aswell

iMarv:
Added behavior for % syntax and also a few comments on why we check for the empty line there

DaelonSuzuka:
Good idea, I'll check it out later today. 

Do you mind taking a look at what happens when you drag the root node of a scene into the editor? I was getting this yesterday: 

Code_JyOLsccST4

Which is definitely not useful.

iMarv:
Will do. I am not too experienced with VsCode extensions - how would we test this automatically? It feels like the edgecases are starting to pile up and I would feel more comfortable with my changes if I would add a few tests on them

DaelonSuzuka:
This extension does have a test suite, but I have no idea how to programmatically invoke behavior like dragging and dropping. 

Personally, I think what we have here is extremely straightforward and doesn't really need testing anyways.

iMarv:
Alright then, not going to start a religious discussion on that 😄 
For me this looks "feature complete" now, I have nothing else to add.

DaelonSuzuka:
Sounds good. I'll review this again later today and probably merge it!

PR #665, opened 2024-06-13
Improve Scene Preview ergonomics

DaelonSuzuka:
"Pinning" a scene in the Scene Preview was  being confused with "pinning" a scene as the debug/launch target.

Renaming it to "locking" and changing the associated icons should help prevent this confusion in the future.

### DIFF
--- a/package.json
+++ b/package.json
@@ -149,7 +149,11 @@
 			{
 				"category": "Godot Tools",
 				"command": "godotTools.scenePreview.refresh",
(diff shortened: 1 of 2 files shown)


DaelonSuzuka:
I added commands for opening the Scene Preview's target scene, and the "main script" of the target scene, if it exists:

Code_2kqHOBjkCC

I also added the existing "refresh scene preview" command as a button.

PR #674, opened 2024-07-01
Fix various syntax highlighting issues

DaelonSuzuka:
Old:

Code_EHbXfjjetl

New:

Code_RQj17cr7tR

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,7 @@
 				"vscode-languageclient": "^7.0.0",
 				"vscode-oniguruma": "^2.0.1",
 				"vscode-textmate": "^9.0.0",
-				"ws": "^8.13.0",
+				"ws": "^8.17.1",
 				"ya-bbcode": "^4.0.0"
 			},
 			"devDependencies": {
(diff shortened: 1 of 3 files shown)


DaelonSuzuka:
I'm currently experiencing a very strange issue:

image

`preload()` is supposed to be dark blue like a keyword, but it's white. It has this `meta.global-method` tag, which is NOT PRESENT in the GDScript grammar:

image

It doesn't happen in my Extension Development Host, it only happens when I actually build and install a `.vsix`:

image

I originally thought I was using some regex rule that works differently in debug vs prod builds, but I created an empty VSCode profile and installed only the `.vsix` build, and the highlighting is correct there. This has to be coming from another extension.

@redCMD do you know how to determine the origin of a particular textmate scope or audit all my installed extensions or something?

RedCMD:
you could try installing the pre-release of my extension
then click the TextMate flower on the sidebar

currently it won't always pick the correct grammar if there multiple with the same scopeName
I'll have to work on that soon™️

which I think is also your issue
prob another extension supplying a grammar with the same scopeName

DaelonSuzuka:
> you could try installing the pre-release of my extension
> then click the TextMate flower on the sidebar

God damn, that's cool! Tracing the call stack found the exact culprit in about 15 seconds! A theme I had installed also had an injection grammar with a couple of extra rules, and the tags came from there.

Thanks for the quick response, you and your extension are an unbelievable help. 

> I think the extension host has priority over other extensions

Lol, I forgot to mention that I run my extension host with a temporary, clean profile to reduce load times and to avoid negative/accidental interactions with other extensions. That's why it worked when debugging but not when I installed the `.vsix` into my main profile.

PR #700, opened 2024-08-20
fix format_documentation for `[code skip-lint]`

funnbot:
ya-bbcode does not consider skip-lint to be an attribute, and it turns into the tag '[code skip-lint]' which won't match with the closing tag `[/code]`, leading to broken formatting.
I don't know if its valid bbcode or not, but Godot docs have it, for example the `Control` class documentation.


Calinou (approved):

DaelonSuzuka:
Hey, thanks for catching that, and congrats on your first contribution.

PR #708, opened 2024-09-04
fix: No data when using typed arrays #699

### LINKED ISSUE #699: The debugger doesn't show any data
### Godot version

4.3

### VS Code version

1.92.2

2.1.0

### System information

macOS Sonoma 14.6.1

### Issue description

The debugger doesn't show any data (locals, members or globals) and watches are also not working. It is also impossible to update the list of watches while the debugger is running.

### Steps to reproduce

- Place a breakpoint in any GDScript
- Run your game until reaching the breakpoint
- The debugger panel in VSCode should be empty (except for the active scene and inspector)

n4nn31355:
Fixes #699 at least in some cases.

The current problem is that for some typed arrays their type is [added to the buffer](https://github.com/godotengine/godot/blob/77dcf97d82cbfe4e4615475fa52ca03da645dbd8/core/io/marshalls.cpp#L1801), but the extension doesn't handle it.

I'm not sure about my solution, but it fixes problem in my case.

Code to reproduce issue:
```python
class_name MainLevel extends Node3D

func _ready() -> void:
    Performance.add_custom_monitor("custom/the/test", _get_calculation_performance)

func _unhandled_input(event: InputEvent) -> void:
    if event.is_action_pressed("ui_cancel"):
        _register_monitor()

func _register_monitor() -> void:
    Performance.add_custom_monitor("custom/the/test_more", _get_calculation_performance)

func _get_calculation_performance() -> float:
    return 0
```

### DIFF
--- a/src/debugger/godot4/variables/variant_decoder.ts
+++ b/src/debugger/godot4/variables/variant_decoder.ts
@@ -23,6 +23,7 @@ import {
 	Projection,
 	ENCODE_FLAG_64,
 	ENCODE_FLAG_OBJECT_AS_ID,
+	ENCODE_FIELD_TYPED_ARRAY_BUILTIN,
 	RID,
 	Callable,
 	Signal,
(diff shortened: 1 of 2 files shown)


NathanNgo:
With Types Dicts coming in soon, do you think it will break on that as well?

DaelonSuzuka:
@n4nn31355 I have good news and bad news:

> I'm not sure about my solution, but it fixes problem in my case.

Good news: Your solution is exactly correct, because it's _almost_ line-for-line identical to what I wrote in #715 
Bad news: I wrote and merged #715 without looking at open PRs, so I accidentally reimplemented your work and stole your thunder. 

Thank you _very_ much for submitting a fix, and I'm terribly sorry for wasting your effort.

Issue #726, opened 2024-09-25
Project is opened with lowercase drive letter


KoBeWi:
### Godot version

4.3, 4.4 dev2

### VS Code version

1.93.1

### Godot Tools VS Code extension version

2.2.0

### System information

Windows 10

### Issue description

When opening project with VS Code, seems like the argument passed to `--path` has lowercase drive letter (e.g. `c:` instead of `C:`). I was not able to 100% confirm it, because I don't know how to check the exact command used to run project. But doing `ProjectSettings.globalize_path("res://")` will return path with lowercase drive.

I originally reported it as Godot bug:
https://github.com/godotengine/godot/issues/97147

https://private-user-images.githubusercontent.com/2223172/369533462-04078e55-af61-459e-b873-ca7e658f5798.mp4

### Steps to reproduce

1. Open workspace with Godot Editor
2. Print project path or use FileDialog
3. Absolute paths will have lowercase drive letter

DaelonSuzuka:
Fixed in #727, but I'm not sure when the next release will be published.

KoBeWi:
Well I'll use the CI build until then.

DaelonSuzuka:
@KoBeWi  FYI the fix for this was just published in `2.3.0`.

PR #727, opened 2024-09-25
Capitalize the drive letter in windows absolute paths

DaelonSuzuka:
RE: #726 

@KoBeWi please grab the CI build and let me know if this solves your problem

### DIFF
--- a/src/utils/project_utils.ts
+++ b/src/utils/project_utils.ts
@@ -33,6 +33,10 @@ export async function get_project_dir(): Promise<string | undefined> {
 	}
 	projectFile = file;
 	projectDir = path.dirname(file);
+    if (os.platform() === "win32") {
+        // capitalize the drive letter in windows absolute paths
+        projectDir = projectDir[0].toUpperCase() + projectDir.slice(1);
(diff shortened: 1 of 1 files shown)


KoBeWi:
It solves the problem, but looks like a hack. Why is `projectDir` like that?

DaelonSuzuka:
I don't understand the question. Why is `projectDir` like what?

DaelonSuzuka:
`path.dirname()` is from the Node standard library, and apparently it returns Windows paths with a lower case drive letter. What is it you're expecting me to do about that?

KoBeWi:
Ah so it's just some Node weirdness.

> What is it you're expecting me to do about that except fix the return value?

I was mostly curious, because it looks like bandaid fix. But if the problem is standard library then it's the only way I guess.

DaelonSuzuka:
> Ah so it's just some Node weirdness.

Pretty much. It preserves the case in the rest of the path, for some reason: ie it doesn't lower the path part of `p:/SkyknightsOnline` on my machine, just the drive letter.

PR #731, opened 2024-10-08
Fix typed arrays of scripts not being decoded properly

mihe:
Related to #699.

This is likely the same issue as what's mentioned [here](https://github.com/godotengine/godot-vscode-plugin/issues/699#issuecomment-2372611126), which seems to stem from the fact that `decode_TypedArray` doesn't take into account all the forms that typed arrays can be encoded in.

Typed arrays can conditionally encode either a `uint32_t` as the type, if the array contains a built-in type, or a `String` if the array contains scripts of some kind, and `decode_TypedArray` currently only accounts for the former and not the latter.

This PR fixes this issue by incorporating not only `HEADER_DATA_FIELD_TYPED_ARRAY_BUILTIN`, but also `HEADER_DATA_FIELD_TYPED_ARRAY_CLASS_NAME` and `HEADER_DATA_FIELD_TYPED_ARRAY_SCRIPT`, as found [here](https://github.com/godotengine/godot/blob/4c4e67334412f73c9deba5e5d29afa8651418af2/core/io/marshalls.cpp#L73-L75).

Similar to what's already being done, this also just discards this new script name/path.

### DIFF
--- a/src/debugger/godot4/variables/variant_decoder.ts
+++ b/src/debugger/godot4/variables/variant_decoder.ts
@@ -23,7 +23,10 @@ import {
 	Projection,
 	ENCODE_FLAG_64,
 	ENCODE_FLAG_OBJECT_AS_ID,
-	ENCODE_FLAG_TYPED_ARRAY,
+	ENCODE_FLAG_TYPED_ARRAY_MASK,
+	ENCODE_FLAG_TYPED_ARRAY_BUILTIN,
+	ENCODE_FLAG_TYPED_ARRAY_CLASS_NAME,
+	ENCODE_FLAG_TYPED_ARRAY_SCRIPT,
 	RID,
 	Callable,
 	Signal,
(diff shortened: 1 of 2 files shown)


DaelonSuzuka:
Hey, thanks for investigating and submitting a fix! Could you share some example GDScript that breaks the existing code so I can test this?

mihe:
Oh, yes, sorry!

It's as simple as declaring a custom resource as `class_name MyClass extends Resource` and then declare a typed array member variable somewhere else as `Array[MyClass]`. Once you hit a breakpoint where that typed array is in scope then you'll see the error mentioned [here](https://github.com/godotengine/godot-vscode-plugin/issues/699#issuecomment-2372611126) and the `Locals`, `Members` and `Globals` lists in the debugger view will all be empty.

DaelonSuzuka:
Awesome, I can confirm that this fixed it:

image

Thanks @mihe!

mihe:
Great to hear!

Any idea when there might be a new release cut with this included?

Issue #735, opened 2024-10-18
Enum members and const variable names should not be tagged as language constants


miss-programgamer:
### Godot version

4.3.stable

### VS Code version

1.93.1

### Godot Tools VS Code extension version

2.3.0

### System information

Windows 10

### Issue description

Currently, enum members and const variables are tagged as `constant.language.gdscript`, which is incorrect. They should be tagged as `variable.other.enummember.gdscript` and `variable.other.constant.gdscript` respectively.

### Steps to reproduce

Run vscode with this extension enabled and write code that includes enum members and/or const variables, I guess.

PR #737, opened 2024-10-18
Fixed the textmate grammar erroneously tagging enum members and const variables as language constants

miss-programgamer:
The title says all. Related to [Issue #735](https://github.com/godotengine/godot-vscode-plugin/issues/735).

### DIFF
--- a/syntaxes/GDScript.tmLanguage.json
+++ b/syntaxes/GDScript.tmLanguage.json
@@ -367,7 +367,7 @@
 			"match": "\\b([A-Z][a-zA-Z_0-9]*)\\.([A-Z_0-9]+)",
 			"captures": {
 				"1": { "name": "entity.name.type.class.gdscript" },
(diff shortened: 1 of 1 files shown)


DaelonSuzuka:
What exactly is the difference between `variable.other.constant.gdscript` and `constant.language.gdscript`? Are they actually highlighted differently?

miss-programgamer:
> What exactly is the difference between `variable.other.constant.gdscript` and `constant.language.gdscript`? Are they actually highlighted differently?

Yes, using the default vscode colors as a frame of reference, language constants are colored with the same dark blue as other keywords, whereas user constants are colored in a lighter electric blue.

Before my changes:
image

After my changes:
image

This is standard in several other languages grammars, like for instance C++ and Python.

Calinou (approved):
This better matches syntax highlighting in other languages such as C++, so it makes sense to me.

PR #738, opened 2024-10-19
Add `@static_unload` annotation and Godot 4.3 Variant types to syntax highlighting

Calinou:
See [Variant class](https://docs.godotengine.org/en/latest/contributing/development/core_and_modules/variant_class.html) for a full list of types available as of Godot 4.3.

- This closes https://github.com/godotengine/godot-vscode-plugin/issues/736.

PR #739, opened 2024-10-19
Remove `OS`, `GDScript` and `Object` from the list of builtins in syntax highlighting

Calinou:
- Follow-up to https://github.com/godotengine/godot-vscode-plugin/pull/738 (can be merged independently, I'll rebase).

OS is a singleton like any other, not a special keyword in GDScript. GDScript is a standard class, not a special keyword either. Object is a Variant type, but it's also a nullable type that is highlighted as a standard object in the Godot script editor.


DaelonSuzuka:
@Calinou this is ready to be rebased. (I'm not smart enough to use the GitHub merge editor)

syntaxes/GDScript.tmLanguage.json
```diff
@@ -437,7 +437,7 @@
...
+			"match": "(?<![^.]\\.|:)\\b(Vector2|Vector2i|Vector3|Vector3i|Vector4|Vector4i|Color|Rect2|Rect2i|Array|Basis|Dictionary|Plane|Quat|RID|Rect3|Transform|Transform2D|Transform3D|AABB|String|Color|NodePath|PoolByteArray|PoolIntArray|PoolRealArray|PoolStringArray|PoolVector2Array|PoolVector3Array|PoolColorArray|bool|int|float|Signal|Callable|StringName|Quaternion|Projection|PackedByteArray|PackedInt32Array|PackedInt64Array|PackedFloat32Array|PackedFloat64Array|PackedStringArray|PackedVector2Array|PackedVector2iArray|PackedVector3Array|PackedVector3iArray|PackedVector4Array|PackedColorArray|super)\\b",
 			"name": "entity.name.type.class.builtin.gdscript"
 		},
 		"const_vars": {
```
AlfishSoftware, line 443: > OS is a singleton like any other, not a special keyword in GDScript.

@Calinou OS is unusual, as it's all uppercase even though it's a type/class. Are you sure it won't be matched by `const_vars` and be colored like a constant (such as NAN, INF)?
If so, maybe it should be added as a special case in `pascal_case_class`.

Also, I don't know if it's just OS that breaks the norm, or if there's others (except RID, which is already listed above as a basic type).

AlfishSoftware: Ah OS is definitely not the only one. On a quick look through the class list in stable, there's also these:
IP, JSON, JSONRPC, UPNP, XRVRS

DaelonSuzuka: Yeah, this was why I added OS to this list in the first place.

Issue #748, opened 2024-11-09
Show non-crashing errors from the debugger


m-radzikowski:
### Godot version

4.3.stable

### VS Code version

1.95.2

### Godot Tools VS Code extension version

2.3.0

### System information

Windows 11

### Problem statement

Some errors do not crash Godot - for example signals handlers with improper number of arguments. Those are only displayed in the Debugger -> Errors tab in the Godot editor:

![image](https://github.com/user-attachments/assets/e1a59063-866e-4a2e-9a55-11de0fcb8a09)

> E 0:00:00:0588   battle.gd:17 @ @selected_unit_setter(): Error calling from signal 'unit_selection_changed' to callable: 'HBoxContainer(hit_chance_bar.gd)::_on_attacker_changed': Method expected 2 arguments, but called with 1.

However, they do not show up in the VS Code, the Debug Console, or the Output -> Godot Debugger.

When you only always ever run from VS Code, you end up not seeing errors that silently break the game.

### Proposed solution

Show the errors from the Godot Debugger -> Errors tab. I wouldn't mind having them in the Debug Console, mixed with my own logs, but at least very visible.

Another option would be to break on those errors, but I don't think this is possible in Godot itself.

Unless, of course, this should already work and it does not for me for some reason?

DaelonSuzuka:
You are absolutely correct, errors are not being routed to the debug console! I can't believe I missed that when I rewrote the debugger (and nobody noticed until now)!

Thanks for the report.

DaelonSuzuka:
Here's what I got so far:

![image](https://github.com/user-attachments/assets/2eb63314-0cb2-472c-a684-c14aa6b5276f)

I would love to find a way to make these collapsible... It's clearly possible because debugging nodejs gets you this:

![image](https://github.com/user-attachments/assets/022da8e9-3781-4aa3-aa2b-5e064aef737f)

kevzettler:
does this also effect use of `push_errors` ? I am not seeing any of those messages in the vscode debug console

DaelonSuzuka:
@kevzettler This PR hasn't been published yet.

PR #749, opened 2024-11-09
Implement warnings and errors in Debug Console

DaelonSuzuka:
#748 noticed that warnings and errors were never hooked up to the debug console.

The items are expandable/collapsible, and the links on the right side of the panel work for any file inside the user's project.

Taskmgr_cUj72S8QmK

Taskmgr_KOKVPwhPtV

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -16,6 +16,10 @@
 			"outFiles": [
 				"${workspaceFolder}/out/**/*.js"
 			],
+			"skipFiles": [
+				"**/extensionHostProcess.js",
+				"<node_internals>/**/*.js"
+			],
 			"preLaunchTask": "npm: watch",
 			"env": {
 				"VSCODE_DEBUG_MODE": true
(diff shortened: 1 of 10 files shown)


Kevin-Jonaitis:
FYI, I was trying to pull in master to pick up this change...unfortunately https://github.com/godotengine/godot-vscode-plugin/pull/731(which is on master but not released yet) seems to break the debugger if you're using typed arrays(I'm running godot 4.3 fwiw). This PR which is currently not merged yet fixes again the typed arrays:
https://github.com/godotengine/godot-vscode-plugin/pull/764

All that to say, master is indeed unstable!

Kevin-Jonaitis:
@DaelonSuzuka apologies I just saw this. The typed _dictionary_ change #764 seemed to fix the typed _arrays_ change #731 that was broken. When I was digging in a bit it seems they touched the same bits of code, which is perhaps why the regression was fixed with a seemingly unrelated change.

Electroshockist:
I, too, am having this problem. It becomes really annoying when trying to use built-in typed arrays like`Area2D.get_overlapping_bodies()`.

The only way i've found to fix this is to append arrays with different data types. Explicitly typing as a generic array doesn't seem to work.

Here are some images:
Screenshot 2025-01-11 201943
Screenshot 2025-01-11 202057

jasoben:
This seems like a good feature, but is there a way to turn it off and return to the nice clean Debug Console output from 2.3.0 (aside from rolling back to 2.3.0, as I've done)? I don't need to see the same random errors I haven't been able to fix yet every time I run my project-- I just want to see my code Output ala the "Output" window in the Godot editor.

PR #752, opened 2024-11-11
Overhaul LSP Client

DaelonSuzuka:
This is a significant refactor of the LSP client code. The overall control flow is dramatically simplified, and this paves the way for future improvements to the LSP client.

@Calinou I'm considering removing websocket support, do you have an opinion on this? The language server switched to a TCP socket in 3.2.2, which was a long time ago now.


Calinou:
> @Calinou I'm considering removing websocket support, do you have an opinion on this? The language server switched to a TCP socket in 3.2.2, which was a long time ago now.

Yes, it should be fine to remove in the next release.

DaelonSuzuka:
@Calinou I did actually spend 20 minutes reimplementing it, but I downloaded both `3.1.2-stable` and `3.2-stable` and couldn't connect to either one via websocket. If anybody actually complains I don't mind revisiting it, but I'm not willing to spend potentially an entire afternoon on something nobody will likely want.

PR #754, opened 2024-11-15
Add newline when dropping nodes into editor

k-expon:
After dropping nodes into editor, it required very little modification. 
So I simply added a newline character at the end of `@onready` generation statements to avoid unnecessary manual line breaks.


DaelonSuzuka:
Did you check the behavior of other "drag into editor" features?

k-expon:
> Oh, that's interesting! Dragging various things into markdown files gives you a snippet-like text insertion with a pre-selection.
> 
> It doesn't add an extra newline, though. I'll try to find some other examples.
> 
> Generally dragging a file into a document gives you the workspace-relative path of the dragged file, again with no newline.

I haven't tested everything related to this, but at least the things you mentioned I've also checked and PR.
As you can see from the modified code, it's only affecting `@onready`

PR #746, opened 2024-11-03
Various Formatter Improvements

DaelonSuzuka:
This introduces a new syntax for defining snapshot tests, and fixes at least #717, #734, #728.

Snapshot tests can now be defined in a single file using "blocks":

```gdscript
# --- IN ---
var  a  =   10
# --- OUT ---
var a = 10
```

I also FINALLY found a solution to remove leading empty newlines from blocks:

```gdscript
# --- IN ---
func test():

	pass
# --- OUT ---
func test():
	pass

# --- IN ---
class Test:

	func _ready():

		pass
# --- OUT ---
class Test:
	func _ready():
		pass
```

PR #756, opened 2024-11-21
Update float syntax rules and formatting

DaelonSuzuka:
Per @AlfishSoftware's [feedback](https://github.com/godotengine/godot-vscode-plugin/pull/746#discussion_r1847281166) or #746, this PR removes the float literal rule I copied from the MagicPython grammar and (hopefully!) fixes the existing float rules.

### DIFF
--- a/syntaxes/GDScript.tmLanguage.json
+++ b/syntaxes/GDScript.tmLanguage.json
@@ -263,21 +263,16 @@
 					"name": "constant.numeric.integer.hexadecimal.gdscript"
 				},
 				{
-					"match": "[-]?([0-9][0-9_]+\\.[0-9_]*(e[\\-\\+]?[0-9_]+)?)",
+					"match": "\\.[0-9][0-9_]*([eE][+-]?[0-9_]+)?",
 					"name": "constant.numeric.float.gdscript"
 				},
 				{
-					"match": "[-]?(\\.[0-9][0-9_]*(e[\\-\\+]?[0-9_]+)?)",
+					"match": "([0-9][0-9_]*)?\\.[0-9_]*([eE][+-]?[0-9_]+)?",
 					"name": "constant.numeric.float.gdscript"
 				},
 				{
-					"match": "[-]?([0-9][0-9_]*e[\\-\\+]?\\[0-9_])",
+					"match": "[0-9][0-9_]*[eE][+-]?[0-9_]+",
 					"name": "constant.numeric.float.gdscript"
-				},
-				{
-					"name": "constant.numeric.float.gdscript",
-					"match": "(?x)\n  (?<! \\w)(?:\n    (?:\n      \\.[0-9](?: _?[0-9] )*\n      |\n      [0-9](?: _?[0-9] )* \\. [0-9](?: _?[0-9] )*\n      |\n      [0-9](?: _?[0-9] )* \\.\n    ) (?: [eE][+-]?[0-9](?: _?[0-9] )* )?\n    |\n    [0-9](?: _?[0-9] )* (?: [eE][+-]?[0-9](?: _?[0-9] )* )\n  )([jJ])?\\b\n",
-					"captures": { "1": { "name": "storage.type.imaginary.number.gdscript" } }
 				},
 				{
 					"match": "[-]?[0-9][0-9_]*",
--- a/syntaxes/GDScript.tmLanguage.json
+++ b/syntaxes/GDScript.tmLanguage.json
@@ -430,7 +430,7 @@
 			]
 		},
 		"annotations": {
-			"match": "(@)(export|export_color_no_alpha|export_custom|export_dir|export_enum|export_exp_easing|export_file|export_flags|export_flags_2d_navigation|export_flags_2d_physics|export_flags_2d_render|export_flags_3d_navigation|export_flags_3d_physics|export_flags_3d_render|export_global_dir|export_global_file|export_multiline|export_node_path|export_placeholder|export_range|export_storage|icon|onready|rpc|tool|warning_ignore|abstract|static_unload)\\b",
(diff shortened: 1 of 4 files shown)


AlfishSoftware:
The best idea would be to take the Godot source as a reference, but sadly it's difficult to interpret the [tokenizer](https://github.com/godotengine/godot/blob/master/modules/gdscript/gdscript_tokenizer.cpp#L674) code into a regex. But looking at the error messages, it disallows multiple adjacent `_` (though it weirdly allows trailing `_`; probably an oversight on their end).

You should probably replace all `[0-9][0-9_]*` with `\d+(?:_\d+)*` -- or  `\d+(?:_\d+)*_?` if you want to be super accurate and allow even a trailing underscore. Also I think line 270 might also match no mantissa digits, like just a `.` or `.e1`.

I tried to come up with one taking GDShader as a base and modifying it. I think these regexes work for float literals:
- `\d+(?:_\d+)*_?[eE][-+]?\d+(?:_\d+)*_?` (no dot, exponent mandatory)
- `(?:(?:\d+(?:_\d+)*_?)?[.]\d+(?:_\d+)*_?|\d+(?:_\d+)*_?[.])(?:[eE][-+]?\d+(?:_\d+)*_?)?`  
    (with dot: digits are mandatory in at least one of both sides, exponent is optional)

You can inspect and test them here (both are joined with a `|`): https://regex101.com/r/6VUN0B/1

Note that I didn't include word boundaries, as I suppose you already have them on all identifiers and keywords.

---
If you use this, remember to do a similar replacement also on the integer literals (binary, decimal and hex) to disallow consecutive `_` on them as well.

DaelonSuzuka:
@AlfishSoftware Great info, thanks!

> Hex: 0x\h+(?:_\h+)*_? (\h is valid in Oniguruma regex, equivalent to [0-9a-fA-F])

Very cool, I've always tried to stick to very basic regex to hopefully avoid implementation specific details, but you've convinced me to actually read the Oniguruma docs.

> Also I think line 270 might also match no mantissa digits, like just a . or .e1.

<details>
<summary>Good catch, it does!</summary>

image

</details>

> The best idea would be to take the Godot source as a reference 
> \<snip>
> disallows multiple adjacent `_`

I'm conflicted about this because, imo, the goal of the textmate grammar is not to strictly enforce the language/interpreter rules. The exact language rules aren't formally specified, and that's a moving target, _and_ I'm not abandoning support to Godot 3-specific syntax. Therefore, the grammar's target _must_ be the union of all the valid syntax of all versions of Godot 3, all versions of Godot 4, plus some margin to allow for mistakes. 

It's also intended to be used in conjunction with the Godot LSP, which will yell about consecutive underscores anyways:

image

AlfishSoftware:
> I'm not abandoning support to Godot 3-specific syntax.

Ah, you're completely right. I didn't test it in Godot 3 at all, just assumed it was the same. Plus, there's no harm in allowing adjacent underscores as it doesn't conflict with any other syntax.
The `_` in Godot 3 seems to be allowed even after `.` and `e`, so those regexes must be right (except for that dot-only case).

> I've always tried to stick to very basic regex to hopefully avoid implementation specific details

You're right, that's actually a good idea, as there's stuff like porting to GitHub syntax, etc. IIRC, there's not that much difference in Oniguruma; what stands out to me is mostly `\h` and `\G` (which it think can be useful in sub-patterns, if I understand correctly).

AlfishSoftware:
There's no harm in allowing it (and the LSP will error anyway), but I think it's also matching cases with just `_` after `e` without a number. E.g.: `.0e_`, `0.e_`, `0.0e_`, `0e_`

syntaxes/GDScript.tmLanguage.json
```diff
@@ -435,7 +430,7 @@
 			]
 		},
 		"annotations": {
-			"match": "(@)(export|export_color_no_alpha|export_custom|export_dir|export_enum|export_exp_easing|export_file|export_flags|export_flags_2d_navigation|export_flags_2d_physics|export_flags_2d_render|export_flags_3d_navigation|export_flags_3d_physics|export_flags_3d_render|export_global_dir|export_global_file|export_multiline|export_node_path|export_placeholder|export_range|export_storage|icon|onready|rpc|tool|warning_ignore|abstract|static_unload)\\b",
+			"match": "(@)(export|export_group|export_color_no_alpha|export_custom|export_dir|export_enum|export_exp_easing|export_file|export_flags|export_flags_2d_navigation|export_flags_2d_physics|export_flags_2d_render|export_flags_3d_navigation|export_flags_3d_physics|export_flags_3d_render|export_global_dir|export_global_file|export_multiline|export_node_path|export_placeholder|export_range|export_storage|icon|onready|rpc|tool|warning_ignore|abstract|static_unload)\\b",
```
AlfishSoftware, line 433: Unless this was supported in Godot 3, I don't think an `@abstract` annotation is supported (yet)?
Looking at https://github.com/godotengine/godot/pull/67777, it's likely going to be a keyword instead of an annotation, if it that PR is merged.

DaelonSuzuka: Thanks @AlfishSoftware, I moved `abstract` from annotations to keywords.

AlfishSoftware:
`._` is being matched by float as well, e.g. in `obj._private_prop`

DaelonSuzuka:
*This* PR is definitely in v2.4.0, but there are several more since, plus there was an open PR for grammar improvements that didn't make it, **plus** I need to do a review of the grammar/formatting issues and see what I've missed. 

Thanks for the reports, I'll make sure to check those cases.

AlfishSoftware:
I see v2.5.0 has fixed the `.` issue.
But in case you're unaware, cases like `.e0` in `obj.e0prop` are still being colored as float constant.

https://github.com/godotengine/godot-vscode-plugin/blob/0da21f23a329cd76dfcf24eb6782ec1173d16fb0/syntaxes/GDScript.tmLanguage.json#L278-L281

DaelonSuzuka:
Thanks for the report @AlfishSoftware. I know there's a few more little problems like this that have been reported but I haven't gotten to yet. I'm gonna go through issues soon and make a list.

PR #764, opened 2024-12-09
Add debugger support for Typed Dictionaries

DaelonSuzuka:
Adding support for Typed Dictionaries is similar to Typed Arrays from #715. 

This is still ONLY protocol decoding support. Carrying the type information through to the debugger UI will require rewriting a lot of code.

@Calinou Do you know when `4.4-stable` is expected to release? I would like to have this published as `v2.4.0` BEFORE the new engine drops, instead of playing catch-up like usual.


Calinou:
> @Calinou Do you know when 4.4-stable is expected to release? I would like to have this published as v2.4.0 BEFORE the new engine drops, instead of playing catch-up like usual.

4.4.beta1 should be sometime this month. Historically, it usually takes at least 2 months for the engine to go from beta to stable, so there's still some time left.

PR #771, opened 2025-01-05
Add Autosave details and Server Port note to README

sabslikesobs:
I've added a few workflow tips to the README. For a workflow that switches between Godot and VSCode, these settings reduce friction, but their existence may not be obvious to someone starting out.


DaelonSuzuka:
Holy crap, I didn't even know about `import_resources_when_unfocused` or `save_on_focus_loss`, those are awesome.

Thanks for these improvements!

Issue #779, opened 2025-01-22
Race condition in godot4\debug_session::get_variable() results in  adding watch and showing variables stops working.


MichaelXt:
### Godot version

v4.3.stable.official [77dcf97d8]

### VS Code version

1.96.4

### Godot Tools VS Code extension version

2.3.0. & Latest repo 2025-01-22, 12:39:49

### System information

Windows 11

### Issue description

![Image](https://github.com/user-attachments/assets/8c1527ce-04d6-48c3-9191-76343a7e1c0f)

# Issue
The debugger watch variables window stops working when the variable is added to watch window.

See repro steps.

# Root cause
There is a race condition between `get_variable()`, `scopes_request()` and the `stack_frame_var`.

### Steps to reproduce

1. Create a godot project with lots of variables. For example generate large Dictionary with lots of keys.
2. launch the godot debugger in vscode. I created a breakpoint in _on_ready
3. Attempt to add any local variable to watch window
  a. Expected: The variable is added correctly
  b. Actual: the variable window shows "=" and no variables added. There is no way to add or remove any more variables. The "Variables" window is also not working.

MichaelXt:
Error as seen in extensions output window:
```log
2025-01-22 14:50:06.375 [error] TypeError: Cannot read properties of undefined (reading 'value')
	at lc.get_variable (c:\Users\Admin\.vscode\extensions\geequlim.godot-tools-2.3.0\out\extension.js:210:68728)
	at lc.evaluateRequest (c:\Users\Admin\.vscode\extensions\geequlim.godot-tools-2.3.0\out\extension.js:210:64742)
```

MichaelXt:
Looking a bit inside the codebase.

It appears that the set_scopes is being called with partial GodotStackVars, where `remaining` is not 0.

## Longer explanation:
1. the `evaluateRequest` sends the request for scopes:
```typescript
await debug.activeDebugSession.customRequest("scopes", { frameId: 0 });
```

2. the `scopesRequest` requests stack_frame_vars and waits for it to complete:
```typescript
await this.got_scope.wait(2000)
```

3. The scopes requests ends in `set_scopes` call with empty `GodotStackVars` with `remaining=15`.

The [set_scopes](https://github.com/godotengine/godot-vscode-plugin/blob/002cfa18a3393636d67ad78aceccfa95e5ffa066/src/debugger/godot4/debug_session.ts#L312) is being called with following GodotStackVars:
```log
[...]

PR #781, opened 2025-01-23
Fix debugger watch window freeze caused by missing responses

MichaelXt:
Fixes the VSCode watch window freeze due to missing response in case of exception. See related #779

### DIFF
--- a/src/debugger/godot3/debug_session.ts
+++ b/src/debugger/godot3/debug_session.ts
@@ -25,7 +25,6 @@ interface Variable {
 	variable: GodotVariable;
 	index: number;
 	object_id: number;
-	error: string;
 }
 
 export class GodotDebugSession extends LoggingDebugSession {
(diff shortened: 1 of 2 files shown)


DaelonSuzuka:
Thanks for submitting a fix, I'll try and check this out tomorrow.

This is issue is honestly not a surprise, I hate that entire section of code. Rewriting it is on my to-do list but I just haven't had time recently.

DaelonSuzuka:
I was able to replicate both the original error, and that these changes fix it. Thanks for contributing!

FYI, when you make a PR like this, it's a bit of an annoyance to the maintainers if you make your changes in the `master` branch on your fork. When I'm reviewing a PR, I typically add the person's fork as a remote and then check out their PR branch, but I couldn't do that here because I already have my own `master` and can't check out yours. I had to manually copy the two files and overwrite my local ones.

PR #786, opened 2025-01-28
Add setting to enable/disable documentation minimap

jnoel-dev:
This adds an option in extension settings to enable/disable documentation minimap. The minimap will be enabled by default.

image

Enabled
image

Disabled
image

### DIFF
--- a/media/docs.css
+++ b/media/docs.css
@@ -1,7 +1,3 @@
-body {
-	margin-right: 200px;
-}
-
 a {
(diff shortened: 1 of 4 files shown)


DaelonSuzuka:
Looks good to me. 

I really wish VSCode's minimap was available as a standalone library. I'm grateful to whoever wrote the library I did use, because it works and I was able to integrate it, but... it's not the best experience, especially on really long documents.

Issue #729, opened 2024-10-03
Debugging: Node type variables show only ID Instead of details


p10tr3k:
### Godot version

v4.3.stable.official [77dcf97d8]

### VS Code version

1.93.1 (Universal)

### Godot Tools VS Code extension version

2.2.0

### System information

MacOs 14.6.1

### Issue description

During debugging in VSCode, when hovering over variables of type Node, I only see the object's ID instead of detailed information about the object. This issue seems to be inconsistent, as I created a copy of the project with a new name and removed the .godot folder, and everything worked correctly again.

However, when I removed the .godot folder in the original project and restarted both Godot and VSCode, the issue persisted.

This may be difficult to reproduce easily, but perhaps someone else has encountered a similar issue and can provide more information.

Expected behavior: Hovering over node variables should display detailed information about the object.
![image](https://github.com/user-attachments/assets/baa6189c-4e7e-4c2a-a078-38e87b6d417a)

Actual behavior: Only the object ID is displayed.
![image](https://github.com/user-attachments/assets/fa316e15-53ae-4201-9698-bee8eb87ea2c)

### Steps to reproduce

[godotbuginstantiateexportsetter.zip](https://github.com/user-attachments/files/17244782/godotbuginstantiateexportsetter.zip)

Steps to reproduce:

Open a project in Godot.
Debug main.gd script containing node type variables.
Set breakpoint on "pass"
Hover over the  'sprite_2d'  variable in VSCode.

MichaelXt:
@DaelonSuzuka this is fixed with latest releases. @p10tr3k do you still see the issue with latest releases?

<img width="1645" height="913" alt="Image" src="https://github.com/user-attachments/assets/34bc75f8-1558-4130-9d38-df74c683a3b1" />

PR #793, opened 2025-02-13
DebugAdapter variables overhaul

MichaelXt:
Fixes multiple issues with variables window for godot 4 projects.
vscode-godot-vars

Note: there is a great initiative of bringing DAP implementation directly into godot debug engine: https://github.com/godotengine/godot/blob/af56d6e8e88e56110577f3bbcfa2eb5cdcc71356/editor/debugger/debug_adapter/debug_adapter_protocol.cpp by @rsubtil and others. We will still need the backward-compatible implementation for now, which also includes remote scene inspector. In future, once the godot server DAP implementation matures, and where will be no need for backward compatibility, we should switch directly to it.

* Redesigned the representation of godot objects to match internal structure of godot server
* Lazy evaluation for the godot objects
* Stack frames now can be switched with variables updated


MichaelXt:
Should fix:
* #729
* #699
* #660
Doesn't fix:
* All variables windows issues for godot 3.4. (Issues opened in 2022 and earlier)

MichaelXt:
* Better presentation for arrays (in-line values, like in vscode javascript DAP)
* Callable and Signal vars are now displayed (though not expandable)

image

MichaelXt:
> Thanks for linking that issue, I haven't been following engine development that closely so I didn't know this was happening.
> 
> Will the DAP implementation support the remote scene inspector?

I'm not sure - just found by looking at the code. @rsubtil should know better.

@DaelonSuzuka, could you review/recommend who can review this PR? Thanks again for maintaining and improving this extension, it is very useful!

DaelonSuzuka:
Yeah I'm gonna review it tomorrow or this weekend.

DaelonSuzuka:
I should know better than to put time estimates into github comments. 

I'll definitely get to it this weekend, though.

rsubtil:
> Will the DAP implementation support the remote scene inspector?

Sorry for the delay; the DAP backend in Godot is basically complete ever since the first implementation, and it didn't have much further work since then. The only really big addition recently was some improvements in object inspection and REPL functionality (https://github.com/godotengine/godot/pull/97585).

The issue with the remote scene inspector is that it is a Godot specific feature, and DAP is agnostic, so there is no way from my understanding to map it into an existing structure. But, you can actually access Godot specific data from DAP, as I've implemented a custom capability precisely for these scenarios. It should still be working (but from my knowledge hasn't been used yet on any plugin, so it definitely needs some testing): https://gist.github.com/rsubtil/01067153db5608d48773719fa63ab2b8#expose-godots-custom-communication

DaelonSuzuka (approved):

PR #794, opened 2025-02-13
Add some useful GDScript snippets for Godot 4

btarg:
Many of the snippets provided by the extension are for Godot 3, and many of them are also not very useful: for instance, there is a snippet to write an "if" statement - which is literally just `if` and then the `condition`. I haven't removed any of the less useful or outdated ones here, but I have provided some that I often use myself to open the discussion on snippets.

I have added the following:
- `void` function shortcut
- `load` and `preload` shortcuts
- Template for a variable with getter and setter (and also a typed version of this)
- Snippet for creating an exported variable in Godot 4
- Snippet for creating a tween
- Snippet for waiting X seconds

### DIFF
--- a/configurations/snippets.json
+++ b/configurations/snippets.json
@@ -1,223 +1,285 @@
 {
-	"Inner class": {
-		"prefix": "class",
-		"body": [
-			"class $1 extends ${2:Reference}",
-			"\t$3"
-		]
-	},
-
-	"Print messages to console": {
-		"prefix": "pr",
-		"body": [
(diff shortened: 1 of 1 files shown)


DaelonSuzuka:
That's awesome, thank you!

btarg:
Thanks for the merge! It feels great to part of the Godot project finally!!! It looks like my last commit to add the @ symbol to the `export` snippet didn't work, I'm still learning about how to make pull requests on GitHub.

PR #797, opened 2025-02-15
Fix #796 - Fixes multi-packet responses breaking things when starting or ending in a multi-byte UTF-8 sequence

### LINKED ISSUE #796: Extension stops working after large LSP response
### Godot version

4.4.beta3, also seems to happen in Godot 4.3.stable

### VS Code version

1.97.1

2.3.0

### System information

Windows 11

### Issue description

On one of my larger projects, the extension seems to stop linting or autocompleting quite frequently. I have to close VSCode and relaunch it to get it to work again. Reloading the window from VS Code's command palette does not fix the issue.

I did some experiments with debugging the extension, and also using a packet sniffer. From what I can tell, it seems like when Godot's LSP sends a large response that spans over multiple TCP packets, the extension immediately tries to parse the first packet as JSON, which produces a JSON.parse error since it isn't the complete response. I could also be entirely wrong, and this is just some sort of buffer copy offset issue.

After the JSON.parse error, everything stops working and VS-Code's devtools console keeps spitting out "Header must provide a Content-Length property." warnings.

VS Code's devlog console:
[vscode-godot-errors.log](https://github.com/user-attachments/files/18811913/vscode-godot-errors.log)

Screenshot of the debugging uncaught exception when this issue happens:
Image

Complete value stored in the msg variable, obtained using `copy(msg)` in the debug console:
[vscode-godot-debug-msg-var.txt](https://github.com/user-attachments/files/18811946/vscode-godot-debug-msg-var.txt)
It looks like msg is only missing the following characters on the end of the string: `}]}`

These characters were absolutely in the TCP packets though:
Image

If there's any more information that I can provide for anyone, please let me know. If I find a solution I'll submit a pull request, but I currently don't expect to get far with my time since I'm unfamiliar with the code base.

### Steps to reproduce

I'm not completely sure other than loading a large Godot project with a lot of gdscript classes. This has been happening on my main project incredibly frequently. Adding GDUnit4 to the Godot project seemed to bloat the LSP responses quite a bit in some situations.

Nibato:
Please see the issue #796 for more details.

### DIFF
--- a/src/lsp/MessageIO.ts
+++ b/src/lsp/MessageIO.ts
@@ -49,7 +49,7 @@ export class MessageIO extends EventEmitter {
 				resolve();
 			});
 			socket.on("data", (chunk: Buffer) => {
-				this.emit("data", chunk.toString());
+				this.emit("data", chunk);
 			});
 			// socket.on("end", this.on_disconnected.bind(this));
 			socket.on("error", () => {


DaelonSuzuka (approved):

DaelonSuzuka:
Thanks again, @Nibato!

PR #800, opened 2025-02-17
Add region block as a language grammar scope

GreenArron:
In Godot's script editor, region blocks and comments are highlighted in different colors for better readability
this tiny PR just adds a region scope to replicate that behavior 
editor_gdscript  
vscode_result_without_override

this also allows regions to be customized as their own separate thing
overriding_color
vscode_result_with_override

I'm not familiar with VSCode grammar scopes so apologies if the scope isn't correct

there's also the option of capturing the name part of the region, but I'm not sure what scope that would have to be


DaelonSuzuka:
This looks fine to me. Thanks for the contribution!

Issue #820, opened 2025-03-11
Request textDocument/documentSymbol failed


devzenfr:
### Godot version

4.4.stable

### VS Code version

1.98.1

### Godot Tools VS Code extension version

2.5.0

### System information

MacOS Sequoia 15.2

### Issue description

Since upgrading to Godot 4.4 and plugin version 2.5.0, opening a gdscript file displays the following error

> [Error - 09:35:45] Request textDocument/documentSymbol failed.
> Error: selectionRange must be contained in fullRange
> 	at tf.validate (file:///Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:108:22635)
> 	at new tf (file:///Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:108:22843)
> 	at Bt (/Users/zoru/.vscode/extensions/geequlim.godot-tools-2.5.0/out/extension.js:35:72393)
> 	at a (/Users/zoru/.vscode/extensions/geequlim.godot-tools-2.5.0/out/extension.js:35:46743)
> 	at Object.GS [as map] (/Users/zoru/.vscode/extensions/geequlim.godot-tools-2.5.0/out/extension.js:35:46829)
> 	at Object.un [as asDocumentSymbols] (/Users/zoru/.vscode/extensions/geequlim.godot-tools-2.5.0/out/extension.js:35:72360)
> 	at c (/Users/zoru/.vscode/extensions/geequlim.godot-tools-2.5.0/out/extension.js:39:56034)
> 	at async QI.provideDocumentSymbols (file:///Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:135:125749) 

The error appeared right after updating plugin to 2.5.0. 

### Steps to reproduce

- Open a Godot 4.4 project. 
- Install Godot vscode plugin
- Open a gdscript file

RobTheFiveNine:
Also having the same issue on 2.5.0 with the following VScode setup:

- **Version**: 1.97.2
- **Commit**: e54c774e0add60467559eb0d1e229c6452cf8447
- **Date**: 2025-02-12T23:20:35.343Z
- **Electron**: 32.2.7
- **ElectronBuildId**: 10982180
- **Chromium**: 128.0.6613.186
- **Node.js**: 20.18.1
[...]

PR #823, opened 2025-03-14
Add ability to suppress LSP error messages

DaelonSuzuka:
Clientside hotfix for #820.

This suppresses the annoying error notification from the failed request.

@Calinou I could add a notification saying that an LSP request failed and suggesting that they move `class_name` to not be at the top of their file? 

image

Something like this?

### DIFF
--- a/src/lsp/GDScriptLanguageClient.ts
+++ b/src/lsp/GDScriptLanguageClient.ts
@@ -3,6 +3,7 @@ import * as path from "node:path";
 import * as vscode from "vscode";
 import {
 	LanguageClient,
+	MessageSignature,
 	type LanguageClientOptions,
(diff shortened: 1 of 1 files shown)


Calinou:
> @Calinou I could add a notification saying that an LSP request failed and suggesting that they move class_name to not be at the top of their file?

I'm torn on this, since the [GDScript style guide](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html) recommends putting `class_name` on the first line of the script.

This means the extension would actively recommend not following the style guide, and going against a formatter if one ends up being implemented in the built-in script editor.

Therefore, I think it's just better to suppress the error for now.

Calinou (approved):

PR #836, opened 2025-03-31
Change debug_current_file error message

testledjones:
Currently, the error message in debug_current_file doesn't tell the user that the scene file and script file must share the same name. This fixes that.


DaelonSuzuka:
Thanks for finding this and submitting a fix!

Issue #838, opened 2025-04-01
Debugger cannot continue after editor Exceptions.


jstraceski:
### Godot version

Godot_v4.4.1-stable_win64

### VS Code version

1.98.2

### Godot Tools VS Code extension version

2.5.1

### System information

Windows 11

### Issue description

When an exception is triggered in an editor script/plugin/tool the debugger becomes unresponsive until the godot editor is closed. i.e. I cannot continue, step over, step into, etc... even after fixing the syntax error.

I am attempting to debug an editor script with this configuration.
```
{
      "name": "GDScript: Launch Project",
      "type": "godot",
      "request": "launch",
      "project": "${workspaceFolder}",
      "debug_collisions": false,
      "debug_paths": false,
      "debug_navigation": false,
      "additional_options": "--editor"
    }
```

When I save my @tool script with a simple syntax issue I am forced to close the editor and reopen.

I see:
```
Exception has occurred.
Parser Error: Expected ":" after dictionary key. (for e.g.)
```

Normally when this happens the script will just fail to compile and I can fix the error and continue but with the debugger active it becomes unresponsive. 
I would be willing to help debug this from the extension side of things as well but I am not familiar with vscode atm.

### Steps to reproduce

1. Open up godot normally
2. Create a new project.
3. Install or create a new plugin. (For example I am using https://godotengine.org/asset-library/asset/713)
4. Close the godot window.
5. Open up vscode with the extension installed.
6. Introduce a syntax error into the plugin somewhere.
7. Add + Execute this launch configuration:
```
{
      "name": "GDScript: Launch Project",
      "type": "godot",
      "request": "launch",
      "project": "${workspaceFolder}",
      "debug_collisions": false,
      "debug_paths": false,
      "debug_navigation": false,
      "additional_options": "--editor"
    }
```

jstraceski:
Just kidding, I made a PR to possibly fix this. https://github.com/godotengine/godot-vscode-plugin/pull/839

DaelonSuzuka:
Yeah that's how they get you.

Calinou:
This was fixed by https://github.com/godotengine/godot-vscode-plugin/pull/839 (as I understand it), closing.

PR #839, opened 2025-04-02
Remove exception guards

jstraceski:
I believe that the additional exception guards are not necessary and in some cases (https://github.com/godotengine/godot-vscode-plugin/issues/838) detrimental. 

Removes the guards in place to prevent the `Continue`, `Next`, `Pause`, `Step In` and `Step Out` debugging actions when encountering an exception. 

Allowing godot to fail appears to be fine in practice.


DaelonSuzuka:
Very interesting. I'll try to check this out tomorrow.

DaelonSuzuka:
Thanks @jstraceski, this looks good to me. I copied your modifications over to the Godot 3 side of the debugger and it looks like it's working.

PR #846, opened 2025-04-11
Fix GDScript Syntax Highlighting for "self" Keyword

YofiAsi:
This pull request updates the GDScript syntax definition to ensure that the keyword "self" is correctly recognized and highlighted. The regex pattern for keywords has been modified to include "self" explicitly, preventing it from being incorrectly matched by other patterns. This enhancement improves the overall syntax highlighting accuracy for GDScript code, ensuring a better development experience.


DaelonSuzuka:
Thanks for your contribution!

Issue #850, opened 2025-04-23
Autoformatter Incorrectly Formats Merge Conflicts


DavidPD:
### Godot version

4.4.1-stable

### VS Code version

1.99.1

### Godot Tools VS Code extension version

2.5.1

### System information

Linux (NixOS Unstable)

### Issue description

When you have git merge conflicts in a GDScript file, the auto-formatter splits the conflict markers causing them to be ignored by merge tools. For instance a merge conflict that looks like this:

```
<<<<<<< HEAD
=======

>>>>>>> f0cd7ef7 (Remove logging and autoformat)
``` 

Will turn into this:

```
<< << << < HEAD
== == == =

>> >> >> > f0cd7ef7(Removelogging and autoformat)
```

Which will cause any git tools to incorrectly assume all conflicts in the file have been fixed. (also there's something funky going on there with that message but that's not the main issue)

## Proposed Solution

It would be awesome if GDScript recognized Git merge conflicts and gave a nice error, but without any assumptions on what someone's using for version control this could be solved by making the auto-formatter skip any lines with a GDScript syntax error instead of _trying_ to format invalid syntax.

Or if it's easier we could just add `<<<<<<`, `>>>>>>`, and `======` to a whitelist or something like that.

### Steps to reproduce

You can either induce a merge conflict using a gdscript file and a couple branches or stashed changes, OR just put the following in any gdscript file and run the auto-formatter (for instance with format on save turned on):

```
<<<<<<< HEAD
=======

>>>>>>> f0cd7ef7 (Remove logging and autoformat)
```

## Expected Behavior

Formatting should not be applied to lines starting with a Git conflict marker.

PR #851, opened 2025-04-23
Fix issue with formatter attempting to format merge conflict markers (#850)

paco3346:


DaelonSuzuka:
Thanks for implementing this. Can you add tests that cover merge conflict markers?

PR #855, opened 2025-04-30
Teach formatter to optionally add two spaces before end-of-line comments.

tmoertel:
This change adds a new formatter option for the number of spaces before an end-of-line comment.

Now you can choose 1 space (the current default) or 2 spaces.

Rationale: Some teams have local style guides that require more than one space to introduce EOL comments. Teams having a Python background, for example, are [accustomed to this style](https://peps.python.org/pep-0008/#inline-comments).

### DIFF
--- a/package.json
+++ b/package.json
@@ -307,6 +307,19 @@
 					"default": false,
 					"description": "Whether extra space should be removed from function parameter lists"
 				},
+				"godotTools.formatter.spacesBeforeEndOfLineComment": {
+					"type": "string",
+					"enum": [
+						"1",
+						"2"
+					],
+					"enumDescriptions": [
+						"1 space before EOL comments # Like this.",
+						"2 spaces before EOL comments  # Like this."
+					],
+					"default": "1",
+					"description": "Number of spaces before an end-of-line comment"
+				},
 				"godotTools.lsp.serverHost": {
 					"type": "string",
 					"default": "127.0.0.1",
--- a/src/formatter/formatter.test.ts
+++ b/src/formatter/formatter.test.ts
@@ -17,6 +17,7 @@ function normalizeLineEndings(str: string) {
 const defaultOptions: FormatterOptions = {
 	maxEmptyLines: 2,
 	denseFunctionParameters: false,
+        spacesBeforeEndOfLineComment: 1
 };
 
 function get_options(folder: fs.Dirent) {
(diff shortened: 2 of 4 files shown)


DaelonSuzuka:
Good suggestion, and thanks for adding the test cases! At first glance everything looks right, but I'll try to review this properly tonight.

tmoertel:
Regarding that failure, it looks like the formatter tests timed out at 2000ms:

>   1) GDScript Formatter Tests
>        Snapshot Test: double_dot_notation.gd:
>      Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/runner/work/godot-vscode-plugin/godot-vscode-plugin/out/formatter/formatter.test.js)

Since the newly added test passes and the `double_dot_notation.gd` test hasn't changed, is the problem just a matter of the formatter suite, as a whole, now taking over 2 seconds? Can we just increase the timeout for that suite? (I'm new to typescript and not sure where the testing framework keeps its tuning knobs. I couldn't find them after hunting for twenty minutes, or I'd try making the change myself.)

DaelonSuzuka:
Yeah the macos test runner fails intermittently, it's a known thing. I've just been rerunning the job.

tmoertel:
It looks like rerunning the macOS job did the trick: all the tests are happy now.

Do you need anything else from me?

Calinou:
Note that it was discussed to add this guideline to the GDScript style guide, but it was ultimately rejected: https://github.com/godotengine/godot-docs/pull/9433

tmoertel:
To clarify, we should note that what was rejected was making two spaces before an inline comment the *recommended* style.

The agreed-upon resolution was to make no recommendation in the style guide for either one or two spaces, leaving the decision as a personal choice. That's why this proposed commit presents this stylistic choice—one space or two—as an option in the formatter. The default is still a single space for compatibility with previous behavior of the formatter.

Is there any objection to letting local teams make their own decisions on this matter?

DaelonSuzuka:
> Is there any objection to letting local teams make their own decisions on this matter?

No there is not.

> Do you need anything else from me?

Nope! This all looks fine, I've just extremely busy.

tmoertel:
Great! Thanks for your work on godot-tools and for looking at this pull request. :-)

DaelonSuzuka:
@tmoertel Thanks for contributing!

I have one minor nitpick (for next time): Please don't develop features on the master branch of your fork! As a reviewer it's easier on me if I can add your fork as a remote and then checkout your feature branch, but if you work in master than I can't do that because I already have a local master branch.

tmoertel:
Thanks for the tip for next time :-)

PR #864, opened 2025-05-24
WIP: Add Node inspector/editor panel in scene preview

patrickjm:
WIP

https://github.com/user-attachments/assets/babfd171-9d46-4540-af55-e110afbe33f1

TODO

- [ ] Revert changes in example godot project
- [ ] General cleanup
- [ ] Add more editor types (enums, vectors, matrices, colors, materials, resources etc)
- [ ] Edit the scene file in memory in an edited vscode tab; don't edit the file directly, then allow user to save?
- [ ] Bug: Some issues syncing state between the node panel and the scene file (including reset button)
- [ ] Bug: Edit field on keystroke/click, not on blur
- [ ] Bug: Missing script exports; only default properties show up
- [ ] Bug: Hide or disable properties that the editor doesn't support atm
- [ ] Bug: Ugly section headers
- [ ] Bug: Key tooltips show bbscript or some kind of markdown instead of rich text for docs
- [ ] Bug: Scene preview on-click should not collapse the tree; on click should be reserved for selecting a node


DaelonSuzuka:
I'm _extremely_ hesitant to add any feature that actually modifies scene files because this extension explicitly supports all versions of Godot back to like 3.4. How can I be confident that this isn't going to do terrible things to Godot 3 scenes, or explode when it encounters some future change to the scene format?

Additionally, I will not accept this PR in it's present state because:
- you've touched way too many files (SceneParser -> SceneHandler rename is unnecessary, for instance)
- you've committed tons of superficial changes like indentation and reordering imports (didn't proofread?)
- the PR is based on the master branch in your fork, which makes it harder for me to check this out and review it. 
- I don't think I'm willing to accept the maintenance burden of an entirely new web framework + build process [1]
- the ui elements don't fit VSCode's style (will they properly reflect the user's theme?)

[1] Did you consider using a framework like https://github.com/vscode-elements/elements ? Just this morning I got this working merely by adding their bundle.js as a media/asset file and loading it with a <script> tag. (I used https://github.com/elelabdev/net-commander/blob/main/src/modules/pingPanel.ts#L283 as a reference)

Issue #844, opened 2025-04-09
LSP Client does not send `textDocument/didOpen` after reconnecting


HolonProduction:
### Godot version

v4.4.1.stable.official [49a5bc7b6]

### VS Code version

Version: 1.98.2

### Godot Tools VS Code extension version

2.5.1

### System information

Fedora 41

### Issue description

After reconnecting the client to a server, the client does not send `textDocument/didOpen` for documents it opened before the disconnect. Since the server instance might have changed (e.g. users restarting the Godot editor) the server is now receiving change notifications against files the client did not open. This is forbidden by LSP.

### Steps to reproduce

- Open Project in Godot editor
- Open VSCode and connect
- Open a file
- Close the Godot editor
- Wait for the automatic connect attempts to finish
- Reopen the Godot project
- Reconnect the client
- Change something in the file -> no `didOpen` is sent for the file but `didChange`

DaelonSuzuka:
Yeah, that makes sense. I'm sure you'll find more similar problems if you keep digging.

I can probably slap together a dirty hack to fix this but the real solution is going to be rewriting the extension's LSP client. The current design has a permanent lifetime for the client when it looks like it should spawn a new client per connection.

Several other issues have been related to this, so I'll add it to my post-it note.

PR #872, opened 2025-06-15
Recreate LSP Client to prevent out of sync state

HolonProduction:
Fixes https://github.com/godotengine/godot-vscode-plugin/issues/844

Over on the issue a rework of the client was mentioned. This PR is not meant to do this. It's a minimal fix to ensure consistent state, since consistent state is a blocker to move forward with certain refactorings and fixes in the server (e.g. https://github.com/godotengine/godot/pull/105236)

(Also added one unrelated comment to a message filter, since it jumped into my eye)

### DIFF
--- a/src/lsp/ClientConnectionManager.ts
+++ b/src/lsp/ClientConnectionManager.ts
@@ -40,8 +40,7 @@ export class ClientConnectionManager {
 	private connectedVersion = "";
 
 	constructor(private context: vscode.ExtensionContext) {
-		this.client = new GDScriptLanguageClient();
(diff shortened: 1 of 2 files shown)


DaelonSuzuka:
I tried doing this a while ago but got a lot of wierd behavior and decided revisit later. I'll try to make time to test this out today.

Thanks!

DaelonSuzuka:
I need an AI that just slaps me in the face with a robot arm if I ever say I'm going to review something within a certain time window.

Sorry for the delay!

PR #877, opened 2025-06-29
Update GDScript syntax highlighter

dalexeev:
* Update annotation list.
* Remove the `abstract` keyword (does not exist in Godot 3 and Godot 4).
* Replace the `void` keyword with the `void` type.
* Add missing `Variant` type.

This is my first PR on this repository, so let me know if I missed anything. Note that I haven't tested this.

Also, I noticed the following code, but I didn't update it because I'm not sure the current approach is correct (lots of built-in functions and types mixed together) and the list is clearly very outdated:

https://github.com/godotengine/godot-vscode-plugin/blob/4bca5d71a6036fa50dc56d00360bc15551c8adc5/src/providers/documentation_builder.ts#L431-L434

### DIFF
--- a/syntaxes/GDScript.tmLanguage.json
+++ b/syntaxes/GDScript.tmLanguage.json
@@ -245,7 +245,7 @@
 			"captures": { "1": { "name": "keyword.control.gdscript" } }
 		},
 		"keywords": {
-			"match": "\\b(?:class|class_name|abstract|is|onready|tool|static|export|as|void|enum|assert|breakpoint|sync|remote|master|puppet|slave|remotesync|mastersync|puppetsync|trait|namespace|super|self)\\b",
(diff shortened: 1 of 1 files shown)


DaelonSuzuka:
I love that every six months somebody comes and requests to change void to a keyword, and every other six months somebody requests to change void to a type.

dalexeev:
I have a good reason:



(However, I also [changed](https://github.com/godotengine/godot/pull/82516) the highlighting in the built-in editor, so I may be biased.)

Calinou (approved):
Looks good to me.

DaelonSuzuka:
I'm a void typist myself, so thank you for your service.

DaelonSuzuka:
Yeah this looks good to me too. Thanks for updating the annotations, that sort of thing is hard for me to keep up with these days.

PR #887, opened 2025-07-16
Add setting to override project directory location

nikitalita:
The plugin currently provides no means of overriding the project directory location; it currently uses the first directory that contains `project.godot` that it finds. This means that if you're in a workspace with multiple `project.godot`s in various subdirectories, your only recourse is to either re-open the workspace scoped to that folder, or just delete those files.

This PR provides the setting `godotTools.workspace.overrideProjectDir`. Now when we search for the project dir, we check the setting with the scope of the current workspace directory (and every subsequent workspace directory if not found), and if we find `project.godot` using that, we use that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -375,6 +375,11 @@
 					"type": "boolean",
 					"default": true,
 					"description": "Whether to enable inlay hints in GDResource (.tscn, .tres, etc) files"
+				},
+				"godotTools.workspace.overrideProjectDir": {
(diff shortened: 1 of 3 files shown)


nikitalita:
@Calinou

Calinou:
@DaelonSuzuka @HolonProduction Any opinions on this? This seems good to me on the surface.

PR #898, opened 2025-08-02
Fix formatting for super and self keywords

### LINKED ISSUE #890: self keyword adds space after it when formatting
### Godot version

v4.5.beta2.official [e1b4101e3]

### VS Code version

Version: 1.102.14746 

2.5.1

### System information

windows 11

### Issue description

when formatting, the self keyword will add a space after it unless there is a . right after it
this only occurs with self and not with other variable names

https://github.com/user-attachments/assets/97cf2237-6a02-436e-b804-c6bbe418c9b6

### Steps to reproduce

type

[].append(self)
var a = self['1']

press format

see

[].append(self )
var a = self ['1']

DaelonSuzuka:
Fixes #890

### DIFF
--- a/src/debugger/godot4/variables/debugger_variables.test.ts
+++ b/src/debugger/godot4/variables/debugger_variables.test.ts
@@ -93,18 +93,18 @@ async function waitForBreakpoint(
 	ctx?: Mocha.Context,
 ): Promise<void> {
 	const t0 = performance.now();
-	console.log(
-		fmt(
-			`Waiting for breakpoint ${breakpoint.location.uri.path}:${breakpoint.location.range.start.line}, enabled: ${breakpoint.enabled}`,
-		),
-	);
+	// console.log(
+	// 	fmt(
+	// 		`Waiting for breakpoint ${breakpoint.location.uri.path}:${breakpoint.location.range.start.line}, enabled: ${breakpoint.enabled}`,
+	// 	),
+	// );
 	const res = await waitForActiveStackItemChange(timeoutMs);
 	const t1 = performance.now();
-	console.log(
-		fmt(
-			`Waiting for breakpoint completed ${breakpoint.location.uri.path}:${breakpoint.location.range.start.line}, enabled: ${breakpoint.enabled}, took ${t1 - t0}ms`,
-		),
-	);
+	// console.log(
+	// 	fmt(
+	// 		`Waiting for breakpoint completed ${breakpoint.location.uri.path}:${breakpoint.location.range.start.line}, enabled: ${breakpoint.enabled}, took ${t1 - t0}ms`,
+	// 	),
+	// );
 	const stackFrames = await getStackFrames();
 	if (
 		stackFrames[0].source.path !== breakpoint.location.uri.fsPath ||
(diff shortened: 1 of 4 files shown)


theXYZT:
I may be reading this wrong, but it doesn't seem like this fixes: `[self] -> [ self ]`.

DaelonSuzuka:
No it doesn't. (when would that even happen?)

theXYZT:
> No it doesn't. (when would that even happen?)

When you use 'self' as a key for a Dictionary.

DaelonSuzuka:
Yes that's obvious but I'm struggling to imagine when I would ever want that.

DaelonSuzuka:
@theXYZT can you think of any other cases?

image

theXYZT:
In your image, is `{ self: self }` supposed to be the formatted text?

theXYZT:
Thinking about it, is there a reason why this can't be the simplification for part of this fix:

```typescript
if (next === "]") {
    return "";
}

if (prev === "[") {
    return "";
}
```

Are there any cases where you want an opening square bracket / closing square bracket to be followed / preceded by a space?

But perhaps this is for a different pull request to experiment with.

DaelonSuzuka:
> In your image, is { self: self } supposed to be the formatted text?

Hm, I just doubled checked and other dict literals are not expanded here. I've been writing mostly ts lately and a similar object literal does have the extra padding.

Just checked python/ruff, and padding is not added to dict literals there.

I have intentionally avoided adding more rules for this syntax because my formatter does not modify indentation at all. A proper formatter would re-indent list/dict literals, collapse small ones onto one line, expand large ones onto multiple lines, etc. Mine is not smart enough to handle that so I've just left them alone.

Let me mess with it a bit more...

DaelonSuzuka:
image

There we go, much better. I removed the special cases entirely and changed it to treat self/super like regular identifiers. Thanks!

src/debugger/godot4/variables/debugger_variables.test.ts
Calinou, line 1: The lines being changed here should probably be removed entirely rather than commented out (since you can find them back using version control).

theXYZT:
Is there anything blocking this from being pushed? I'd love to have a nightly build with this.

Calinou:
@DaelonSuzuka Is this good to merge now? It appears this has regressed in 2.6.0: https://github.com/godotengine/godot-vscode-plugin/issues/965

Calinou (approved):
Thanks!

jdbool:
Are there plans for a release any time soon? This bug is super annoying. @DaelonSuzuka

DaelonSuzuka:
Yeah as soon as I get caught up on the 9 months of open issues with the 1-3 hours a week I can spare to work on this while trying to recover from being completely burnt out. Should be any day now.

jdbool:
> Yeah as soon as I get caught up on the 9 months of open issues with the 1-3 hours a week I can spare to work on this while trying to recover from being completely burnt out. Should be any day now.

Surely the whole backlog doesn't need to be fixed to cut a new release. A lot of people are probably noticing this bug in their diffs.

Calinou:
See this comment for the next release:

- https://github.com/godotengine/godot-vscode-plugin/issues/972#issuecomment-4383165759

Issue #502, opened 2023-10-11
Editor Path belongs in user settings, not workspace


lalomartins:
### Godot version

4.1

### VS Code version

1.83

### Godot Tools VS Code extension version

1.3.1

### System information

Windows 11

### Problem statement

Spin off from #420 which seems to be fixed now. However, when you set the editor path by using the file picker, I feel it should save to user settings, or at least ask the user where to save. The workspace settings should be committed to version control, and if we're developing in a team, there's good odds each team member will have their Godot editor installed somewhere different. Even just on my own, I have Godot in different places on my desktop and laptop.

### Proposed solution

Save the editor path in user settings when it's input by the extension file picker.

DaelonSuzuka:
> The workspace settings should be committed to version control

This is definitely not universally accepted.

> ask the user where to save

This is a good idea. I'll see if there's a clean way to ask the user what they want, and then look at standardizing all the file picker/settings prompts to do the right thing.

DaelonSuzuka:
@lalomartins 

I also forgot to mention, it's pretty trivial to add the folder with your godot executable(s) to the system path, and then your `editorPath` can just be set to something like `godot4.exe`. This is what I do for extension development when I need to switch between multiple Godot versions easily.

tsanyqudsi:
@DaelonSuzuka I think the User Settings are there now ? Shouldn't this be closed ?

tsanyqudsi:
@DaelonSuzuka I was talking about this.

![image](https://github.com/godotengine/godot-vscode-plugin/assets/1050880/c6b35a02-bde1-4670-b68a-1b1f6fe8a7c9)

Although file picker is a nice addition, but considering that someone that working with Godot should be able to get the file path easily and copy-pasted / type it to the field.

I think it's better to update the `Readme.md` instead:

# Getting Started

You can set VSCode as your default script editor for Godot by following these steps:

## 1. Setup Godot Tools in VSCode
1. Download and install Godot Tools, there are few options to download and install the extension:
- Find **Godot Tools** in Extension section of VSCode(recommended).
- [Visual Studio Marketplace (also recommended)](https://marketplace.visualstudio.com/items?itemName=geequlim.godot-tools): Stable release, with support for automatic updates.
- [GitHub Releases](https://github.com/godotengine/godot-vscode-plugin/releases): Stable release, but no automatic updates. Can be useful if you need to install an older version of the extension.
[...]

PR #905, opened 2025-08-17
Added the ability to specify platform-specific editorPath

ogapo:
- support config of the form `godotTools.editorPath.(windows|macos|linux)-(x86_64|arm64).godot(3|4)`
- match godot's buildsystem terms for familiarity
- move all path manipulation hacks into clean_godot_path
- tried to make sure all code that needs the editor path consistently use the util functions

BUGFIX: expand ${env:X} into the string. Previous code replaced the whole string if it contained an env tag and only matched the first one.

Per some discussion on #502 I realized this is the only thing stopping folks from using workspace config (committed to source control) for configuring prebuilt binaries in their projects given the paths are already treated as workspace relative (if they're not absolute).

### DIFF
--- a/package.json
+++ b/package.json
@@ -272,12 +272,42 @@
 				"godotTools.editorPath.godot3": {
 					"type": "string",
 					"default": "godot3",
-					"description": "Path to the Godot 3 editor executable. Supports environment variables using '${env:VAR_NAME}'."
+					"description": "Path to the Godot 3 editor executable. Workspace relative. Supports environment variables using '${env:VAR_NAME}'."
 				},
 				"godotTools.editorPath.godot4": {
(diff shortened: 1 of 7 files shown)


DaelonSuzuka:
FYI I'm out of town until next week but I'll review this when I get back.

ogapo:
@DaelonSuzuka wondering if you have any cycles to have a look at this? Thanks!

ogapo:
@DaelonSuzuka or @Calinou wondering if there's any chance you could look at this PR? I think it's pretty much g2g

DaelonSuzuka:
I've been extremely busy (new job) but I'll try to look at it this week.

PR #908, opened 2025-08-20
Recognize pascal-case identifiers that ends with 2 or more upper case letters as `pascal_case_class`

### LINKED ISSUE #907: Won't highlight pascal-case class names with 2+ upper case at the end
### Godot version

4.4.1.stable

### VS Code version

1.103.1

2.5.1

### System information

Windows 11

### Issue description

Afaik this only happens when an expression involves autoloads, where the convention is to give the autoload object a name that mimics a class name. Basically you would name a class like "BaseUI", "StateSyncRPC", "MonsterAI", and not "BaseUi", "StateSyncRpc", "MonsterAi" (which highlights correctly).

I think it makes sense the above examples would all be recognized as class names.

### Steps to reproduce

1. Create new .gd file like a "base_ui.gd"
2. Register the new file as autoload, give the object a name like you would for such cases: "BaseUI" (not the default "BaseUi")
3. Call some method on this autoload, like: `BaseUI.add_child(Node.new())`
4. Produces this coloring: <br />

lfod1997:
Fixes #907 

Before:

After:

### DIFF
--- a/syntaxes/GDScript.tmLanguage.json
+++ b/syntaxes/GDScript.tmLanguage.json
@@ -454,7 +454,7 @@
 			"name": "variable.other.constant.gdscript"
 		},
 		"pascal_case_class": {
-			"match": "\\b([A-Z]+[a-z_0-9]*([A-Z]?[a-z_0-9]+)*[A-Z]?)\\b",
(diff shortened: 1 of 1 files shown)


DaelonSuzuka:
That makes sense to me and the grammar change looks correct. Thanks!

PR #910, opened 2025-08-22
Use cached values for project version and directory

### LINKED ISSUE #645: v2.0.0 maxes out my CPU
### Godot version

4.2.2-rc2

### VS Code version

1.88.1

2.0.0

### System information

Windows 11

### Issue description

Hard to explain why exactly, but after coding for a while, the extension causes 100% CPU usage on my desktop PC. Has happened on other devices as well. As soon as I disable the extension the CPU usage drops to normal amounts.

The plugin must be connected to Godot for this to occur. If the plugin disconnects (after Godot has closed), the CPU usage will persist until all 10 connection trials have completed. Reconnecting afterwards does not yield anymore whacky CPU usage.

Again, I wish I could give more info on this, but I am still trying to diagnose it.

image

### Steps to reproduce

I wish I could give more info, all I did was code for an hour or so, and eventually my computer became extremely slow and I began trying to find the issue.

fehnomenal:
Fixes #645 

The performance issue was caused by scanning all folders on every call because the project directory was cached but not reused.

### DIFF
--- a/src/utils/godot_utils.ts
+++ b/src/utils/godot_utils.ts
@@ -19,6 +19,10 @@ let projectDir: string | undefined = undefined;
 let projectFile: string | undefined = undefined;
 
 export async function get_project_dir(): Promise<string | undefined> {
+	if (projectDir && projectFile) {
+		return projectDir;
+	}
+
 	let file = "";
 	if (vscode.workspace.workspaceFolders !== undefined) {
 		const files = await vscode.workspace.findFiles("**/project.godot", null);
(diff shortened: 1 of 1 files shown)


DaelonSuzuka:
Dang, I was 75% sure this would be related to scene file parsing or the custom document links provider.

Yeah this looks like a reasonable fix, thanks!

(This will *probably* have a cache invalidation issue, but this entire section is due for a rewrite anyways. I'm going to merge this and deal with the chance of a problem later.)

PR #912, opened 2025-08-24
Make sure last LSP client is disconnected before creating a new one. 

jstraceski:
The LSP client seems to have left over resources when restarting godot. 
This causes CTRL-hover to fail after restarting godot because multiple `textDocument/definition` requests are sent.

Steps to reproduce issue:
1. Open vscode extension.
1. Open godot project.
1. Within the vscode window CTRL-Click any object.
1. Close + reopen godot.
1. Within the vscode window CTRL-Click any object.

If you have a debuging window open you should see `rx`/`tx` similar to this:
```
[lsp.io] tx: {jsonrpc: '2.0', id: 4, method: 'textDocument/hover', params: {…}}
logger.ts:70
[lsp.io] tx: {jsonrpc: '2.0', id: 22, method: 'textDocument/hover', params: {…}}
logger.ts:70
[lsp.io] rx: {id: 4, jsonrpc: '2.0', result: {…}}
logger.ts:70
[lsp.io] tx: {jsonrpc: '2.0', method: '$/cancelRequest', params: {…}}
logger.ts:70
[lsp.io] tx: {jsonrpc: '2.0', id: 5, method: 'textDocument/definition', params: {…}}
logger.ts:70
[lsp.io] tx: {jsonrpc: '2.0', id: 23, method: 'textDocument/definition', params: {…}}
logger.ts:70
[lsp.io] tx: {jsonrpc: '2.0', id: 6, method: 'textDocument/hover', params: {…}}
logger.ts:70
[lsp.io] tx: {jsonrpc: '2.0', method: '$/cancelRequest', params: {…}}
logger.ts:70
```

This change disconnects io functions, and stops the last valid LSP client before making a new client.

(I did test this on godot > 4.5 but I have made some local changes so there is a chance that it is not required under the current stable godot version.)

### DIFF
--- a/src/lsp/ClientConnectionManager.ts
+++ b/src/lsp/ClientConnectionManager.ts
@@ -75,6 +75,8 @@ export class ClientConnectionManager {
 
 	private create_new_client() {
 		const port = this.client?.port ?? -1;
+		this.client?.io?.removeAllListeners();
+		this.client?.stop();
 		this.client?.events?.removeAllListeners();
 		this.client = new GDScriptLanguageClient();
 		this.client.port = port;


HolonProduction:
Could you check whether your original issue was solved by https://github.com/godotengine/godot-vscode-plugin/pull/968

This PR looks alright as safeguard. But in concept `create_new_client` should only be called when the old client is in a state were it is disconnected and will never reconnect.

Nallebeorn:
I tested a build of this branch, and it seems to fix https://github.com/godotengine/godot-vscode-plugin/issues/978. At least with the repro steps of restarting Godot that are described in the comments.

Calinou:
@HolonProduction Do you think this is safe to merge? If so, I can merge it and tag 2.6.2.

HolonProduction:
The `client?.stop` part makes sense to me. While client `client.io` is public, it really should be a private member and cleaning it up at this place breaks encapsulation to me.

I verified that just the `client?.stop` part fixes https://github.com/godotengine/godot-vscode-plugin/issues/978 it probably contains some code to unregister the client with VSCode.

Calinou:
I've rebased the PR on top of `master` and squashed commits, will merge once CI passes.

PR #913, opened 2025-08-25
Draft: Quickfix extra quotes by removing them from LSP response

### LINKED ISSUE #885: load/preload autocomplete injects extra wrapping "
### Godot version

4.4

### VS Code version

1.101.2

2.5.1

### System information

Windows 10

### Issue description

when using `load` or `autoload` you get autocomplete hints after you add `""` as the first argument

Image

when you accept one of the autocomplete options, the string is inserted as a string inside of the `""` that was entered to initiate the first argument

Image

I then have to manually clean out the extra `"`. The autocomplete select should insert the string content, instead of inserting a string literal into the enclosing `"`

This issue has been present for as long as I have been using the vscode extension since around the 4.0 release, I've only just now realized how to articulate what I was experiencing.

### Steps to reproduce

Open any `.gd` file in any project, begin creating a variable of any name and typing `load()` or `preload()` as the assignment value of the variable.

Type `"` to initiate the autocomplete hints

Select one of the auto-complete options

the auto-complete will insert the enclosing `"` instead of just completing the string

nick-loginov:
Fixes #885

### DIFF
--- a/src/lsp/GDScriptLanguageClient.ts
+++ b/src/lsp/GDScriptLanguageClient.ts
@@ -14,6 +14,7 @@ import {
 import { globals } from "../extension";
 import { createLogger, get_configuration, get_project_dir } from "../utils";
 import { MessageIO } from "./MessageIO";
+import { LSPMiddleware } from "./LSPMiddleware";
 
 const log = createLogger("lsp.client", { output: "Godot LSP" });
 
@@ -116,6 +117,7 @@ export default class GDScriptLanguageClient extends LanguageClient {
 				{ scheme: "file", language: "gdscript" },
 				{ scheme: "untitled", language: "gdscript" },
 			],
+			middleware: new LSPMiddleware()
 		};
 
 		super("GDScriptLanguageClient", serverOptions, clientOptions);
--- a/src/lsp/LSPMiddleware.ts
+++ b/src/lsp/LSPMiddleware.ts
@@ -1,0 +1,75 @@
+
+
+import * as vscode from "vscode";
+import { Middleware, ProvideCompletionItemsSignature } from "vscode-languageclient";
+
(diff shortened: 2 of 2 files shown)


DaelonSuzuka:
I WAS going to tell you I've tried this by rewriting the incoming messages and it was so unreliable that I'm not interested in a client side bandaid fix anymore, but I've never seen this middleware mechanism. 

I'm very interested to test this out.

nick-loginov:
Yeah, I'm still testing it right now, and you're right, it's all junky and there is more stuff to make it work as intended. In current state only works for simple cases.

DaelonSuzuka:
No problem. We can also slap a setting in front of this and mark it experimental. 

I have some other dirty hacks that I think would be cleaner if I move them to this middleware layer, so this PR won't be wasted even if the quote thing doesn't work out.

nick-loginov:
@DaelonSuzuka I was going to ask what exactly was the problem in your attempts? Did some of the cases not work? Can you give some examples? I'm convinced it's possible to make it reliable

DaelonSuzuka:
That was a looong time ago now so I don't remember the specifics. I could get it working in like, one specific test scenario but it broke when I tried to generalize it.

HolonProduction:
I'd prefer to fix this properly in the server so I started working on that https://github.com/godotengine/godot/pull/117710

But if you want to add a workaround for it anyway feel free to do so, just wanted to cross link this.

DaelonSuzuka:
@HolonProduction I fully agree, (and thanks for fixing this!)

HolonProduction:
Closing due to the upstream fix being merged. If you still find edge cases which don't work correctly feel free to open an upstream issue and we can address it on the server side.

PR #914, opened 2025-08-27
textmate rules: reclassify true/false/null as literal constants 

yotam-frid:
This branch marks the `true`/`false`/`null` consts as language literal constants to mimic the behavior of the engine editor which treats them as such.

Before: both `true`/`false` and `Vector2.RIGHT` treated as same level
image

After: booleans are in a separate scope (correct behavior)
image

### DIFF
--- a/src/formatter/textmate.ts
+++ b/src/formatter/textmate.ts
@@ -113,7 +113,7 @@ function parse_token(token: Token) {
 		token.type = "keyword";
 		return;
 	}
-	if (token.scopes.includes("constant.language.gdscript")) {
+	if (token.scopes.includes("constant.language.gdscript") || token.scopes.includes("constant.language.literal.gdscript")) {
 		token.type = "constant";
 		return;
 	}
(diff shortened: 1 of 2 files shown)


DaelonSuzuka:
What are your screenshots supposed to be demonstrating?

I'm fairly certain that true/false/null are being categorized the way I want them to be, so you're going to need to a better reason than this.

yotam-frid:
> What are your screenshots supposed to be demonstrating?
> 
> I'm fairly certain that true/false/null are being categorized the way I want them to be, so you're going to need to a better reason than this.

Currently, true/false/null are treated as any normal constant. Regular consts are user-definable, but using true/false/null as  a constant name is a no-op in GDScript, since they are reserved language keywords.

This PR doesn't change the classification from "constant", just more specificity: "constant literal". This creates a more accurate classification that aligns behavior with the default Godot script editor, which treats these constants as _keywords_, not constants:
image
(this is inside Godot, not vscode)

In the original PR screenshots (that are from inside vscode), note how `Vector2.RIGHT` and `true` or `false` are colored differently. This was not possible before.

ScatteredComet:
I think this is a great change; it just aligns the vscode editor to the default godot editor. Is there anything else that needs to be done here?

PR #915, opened 2025-08-27
Debugger: fix inspecting objects/variables in Godot v4+

yotam-frid:
In Godot 4, the `scene:inspect_object` remote debugger command was replaced with a multi-object inspect `scene:inspect_objects` command.

The extension only worked with the v3 schema. This branch updates the v4 server controller to work with the new command.

This fixes operations like hover-inspecting variables, or seeing an object's members, which would hang indefinitely before.

### DIFF
--- a/src/debugger/godot4/server_controller.ts
+++ b/src/debugger/godot4/server_controller.ts
@@ -120,7 +120,7 @@ export class ServerController {
 	}
 
 	public request_inspect_object(object_id: bigint) {
-		this.send_command("scene:inspect_object", [object_id]);
+		this.send_command("scene:inspect_objects", [[object_id], false]);
 	}
 
 	public request_scene_tree() {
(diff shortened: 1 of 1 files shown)


DaelonSuzuka:
What version specifically was this change made? I feel like this is a recent change and was not present in 4.0.

yotam-frid:
> What version specifically was this change made? I feel like this is a recent change and was not present in 4.0.

You're right, it was only merged 5 months ago and was added to the 4.5dev1 upstream:
https://github.com/godotengine/godot/commit/5c66129e621e51bd85eef6521ebb775217dfa752

The proposal's been around since [last year](https://github.com/godotengine/godot/pull/99680). Guess it was missed it was a breaking change.

Looks like it's going to be part of version 4.5. So it would be wiser to either add a backwards-compatible implementation, or re-add the old command in the godotengine repo

yotam-frid:
@DaelonSuzuka  I created an issue https://github.com/godotengine/godot/issues/110042, let's see how it's handled.

yotam-frid:
I created a backwards compatibility layer fix - https://github.com/godotengine/godot/pull/110043
If merged, PR can be closed. Otherwise, a forking logic should be added.

DaelonSuzuka:
There's no reason to create an issue on the engine, they can change the APIs if they want to. However, this extension explicitly supports all Godot versions after maybe 3.2, so this PR isn't usable as-is.

Detecting the connected version and forking the internal logic is the correct solution.

yotam-frid:
> There's no reason to create an issue on the engine, they can change the APIs if they want to. However, this extension explicitly supports all Godot versions after maybe 3.2, so this PR isn't usable as-is.
> 
> Detecting the connected version and forking the internal logic is the correct solution.

OK - I did just that 🚀

MichaelXt:
What are the repro steps? I see inspect_object working with extension v2.6.1 and Godot_v4.6-stable
image

PR #848, opened 2025-04-20
Debugger Tool Improvements

DaelonSuzuka:
This PR makes several improvements to the debugger tools.

PR #936, opened 2025-10-23
Stale variables fix

### LINKED ISSUE #935: Debugger regression: variables are not refreshed on callstack change
### Godot version

*

### VS Code version

*

*

### System information

*

### Issue description

Variables are no longer refreshed on callstack change.

This is a regression introduced by #848

### Steps to reproduce

set 2 breakpoints, observe the behavior:
First breakpoint hit:

Image

Second breakpoint hit:

Image

Observe that variables are not refreshed.

MichaelXt:
* Introduced by #848
* Fixes #935

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -18,7 +18,7 @@ body:
     description: >
       Specify the Git commit hash if using a development or non-official build.
       If you use a custom build, please test if your issue is reproducible in official builds too.
-    placeholder: 4.3.stable, 4.4.dev1 (28a72fa43)
+    placeholder: 4.5.1.stable, 4.6.dev0 (6fd949a)
   validations:
     required: true
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,15 +21,15 @@ jobs:
       - name: Install Godot (Ubuntu)
         if: matrix.os == 'ubuntu-latest'
         run: |
-          wget https://github.com/godotengine/godot/releases/download/4.3-stable/Godot_v4.3-stable_linux.x86_64.zip
-          unzip Godot_v4.3-stable_linux.x86_64.zip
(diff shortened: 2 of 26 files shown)


MichaelXt:
@DaelonSuzuka, does this change look good to you?

MichaelXt:
@Calinou 
2.6.1 is released with regression this PR fixes. (Introduced by #848) 
image
notice that the Locals variables are not refreshed:
image

HolonProduction:
On a completely unrelated note:
- Why does this plugin implement the Godot specific debugging protocol, when Godot does support DAP? Is this just a grew over time thing @DaelonSuzuka

I don't think I can give qualified input on this one, I have no clue about the debugging stuff. But given that https://github.com/godotengine/godot-vscode-plugin/pull/848 directly changed that part, I'm a bit worried that this is undermining some caching effort, but it might as well be a harmless mistake. Sadly there isn't any description on the PR so I don't know.

Edit: For context, from my testing the `debug_enter` message occurs for every single step done. So if the goal is to somehow cache stuff between steps, this PR would invalidate that caching. But I can't really think of something that's worth caching in this context, because every step might change basically everything. So my gut feeling tells me this _should_ be fine. But I'm absolutely not sure about it.

DaelonSuzuka:
The godot debug protocol is older than dap has even existed, and it supports godot specific features that dap doesn't.

MichaelXt:
> For context, from my testing the `debug_enter` message occurs for every single step done. So if the goal is to somehow cache stuff between steps, this PR would invalidate that caching. But I can't really think of something that's worth caching in this context, because every step might change basically everything. So my gut feeling tells me this _should_ be fine. But I'm absolutely not sure about it.

@HolonProduction , that is a correct observation. I authored VariablesManager as part of #793 and it is not designed to be reused across `debug_enter` calls.

DaelonSuzuka:
@MichaelXt I updated this branch to master and the diff appears to be empty now. Was this change already implemented in #994 or did I break something?

MichaelXt:
Correct, #944 has that fix too. Thanks!

PR #955, opened 2025-12-28
Replace window reload with restart extension host

WebFreak001:
UI can stay, this is just a faster way to reload extensions without losing unsaved data and has been available for at least 1-2 years in vscode now.

You can try out what this does in the command palette: Ctrl-Shift-P -> Restart Extension Host


DaelonSuzuka:
Thanks, that's a good idea.

Changing settings should honestly never require reloading anything, I just haven't gotten around to finding every place that loads settings and making sure they have a "config changed" handler.

Issue #954, opened 2025-12-21
Some snippets don't have static typing


iker-sr:
### Godot version

4.5.1-release

### VS Code version

1.107.1

### Godot Tools VS Code extension version

2.5.1

### System information

Windows 10

### Issue description

some code snippets don't have static typing, for example `process`, `physics`, `input`.

[configurations/snippets.json](https://github.com/godotengine/godot-vscode-plugin/blob/master/configurations/snippets.json), line 29:
```json
	"_process method of Node": {
		"prefix": "process",
		"body": [
			"func _process(delta):",
			"\t${1:pass}"
		]
	},
	"_physics_process method of Node": {
		"prefix": "physics",
		"body": [
			"func _physics_process(delta):",
			"\t${1:pass}"
		]
	},
	"_input method of Node": {
		"prefix": "input",
		"body": [
			"func _input(event):",
			"\t${1:pass}"
		]
	},
```

### Steps to reproduce

Write `process`, `physics` or `input` in a gdscript file and press enter.

BenBlueeeee:
I believe ready also doesn't?

As a workaround, just being able to disable the snippets would work (then the user can define their own)

ScatteredComet:
Workaround for now:

- Control+p -> select "insert snippet"
- Hover over the snippets that aren't statically typed, and then click the eye icon to hide that snippet from intellisense
- Define our own statically typed snippets for ready, physics_process, etc in User/snippets/gdscript.json

(for a video, see [here](https://github.com/microsoft/vscode/issues/10565#issuecomment-721832613))

IMO, it would be nice for users to be able to disable these snippets. I believe the recommendation is to provide snippets as a seperate extension (as vscode doesn't seem to provide another option to remove snippets from extensions other than the steps I've provided above). (But that's not what this issue is about, so feel free to ignore this)

ScatteredComet:
I've just created a PR that statically types all of the functions listed in snippets.json:

https://github.com/godotengine/godot-vscode-plugin/pull/958

As a side node, some of the included functions provided seem a little arbitrary (for example, _input_event and _draw), but I've left them in.

OmnesPotens:
@iker-sr @BenBlueeeee Should this be closed?

Calinou:
- Fixed by https://github.com/godotengine/godot-vscode-plugin/pull/958.

PR #958, opened 2026-01-07
feat: statically type all function snippets

ScatteredComet:
I have also updated snippet names of the type "[name] method of [class]" to correctly reference the class that defines them.

For example, _draw() was labelled as "of Node" before, but its actually CanvasItem which defines this function.


ScatteredComet:
I just realised that the snippets also dont provide the static typing for their parameters

I'm happy to fix this too I'll add some changes

ScatteredComet:
Should be all done now

ScatteredComet:
Just thought I'd link to the original issue: this addresses https://github.com/godotengine/godot-vscode-plugin/issues/954

ScatteredComet:
Okay that's added in now; I also renamed the inpute prefix to inpute2D and inpute3D. Let me know if these should be changed / if there's some convention that needs to be followed. I wasn't sure if having them both called inpute would be confusing.

Calinou (approved):
Thanks!

Issue #959, opened 2026-01-14
Can't Shift+Drag from Scene Preview on Archlinux


Akira-Kuru:
### Godot version

4.5.1 and 4.5.1-mono

### VS Code version

1.108.0

### Godot Tools VS Code extension version

2.5.1

### System information

Archlinux

### Issue description

As the description explains, I tried using godot+vscode with the godot tools extension on archlinux.
I attempt using godot and vscode over flatpak, AUR and pacman but none of these worked when trying to drag variables from the scene preview with shift.

### Steps to reproduce

1. Have a Archlinux System or setup Archlinux
2. Install Godot / Godot Mono over flatpak, aur or pacman
3. Install VSCode over flatpak, aur or pacman
4. Download & Install Godot-tools Extension
5. Create empty project & setup project to open code with vscode (Follow the Godot Extension Description)
6. Add empty nodes and name them to your choice
7. Create a script.gd
8. Open it in VSCode
9. Drag one of the nodes from the scene preview to script.gd while holding shift

andreygrigol:
I was having the same issue as you. You're probably running Arch on wayland mode, and I figured that's the problem. When I launched VSCode on X11 mode, it worked just fine. Not sure how to fix it in the code, but a workaround is to run code with --ozone-platform=x11 flag and it will work just fine

Akira-Kuru:
> I was having the same issue as you. You're probably running Arch on wayland mode, and I figured that's the problem. When I launched VSCode on X11 mode, it worked just fine. Not sure how to fix it in the code, but a workaround is to run code with --ozone-platform=x11 flag and it will work just fine

Jup you got that on point and I also have an instance which I use Hyprland so wayland is my general go too. That `--ozone-platform=x11` is a good work around 👍

Calinou:
This is a limitation of Wayland support in VS Code. We should probably document this somewhere in the README.

HolonProduction:
This seems to be documented now. Should the issue be closed?

Akira-Kuru:
> This seems to be documented now. Should the issue be closed?

I would like to add my further findings before it gets closed.

This issue is related to Electron (Visual Studio runs on Electron) causing the issues on Linux related Issue here: https://github.com/microsoft/vscode/issues/156723 .
For Archlinux there is a specific flag which can be found here: https://wiki.archlinux.org/title/Wayland#Command_line_flags

Maybe this can be extended into the documentation?
Then we can close the issue.

PR #967, opened 2026-01-29
Added information to README about Linux and Wayland support

andreygrigol:
As related in the issue #959, drag + shift drop doesn't work in Linux systems which are using Wayland protocol as the display server. That's a problem related to VSCode and Wayland compatibility. Added information to the README file so other people can find a workaround to fix the issue, although it would be best to debug and fix the problem with a patch.

### DIFF
--- a/src/lsp/MessageIO.ts
+++ b/src/lsp/MessageIO.ts
@@ -51,10 +51,8 @@ export class MessageIO extends EventEmitter {
 			socket.on("data", (chunk: Buffer) => {
 				this.emit("data", chunk);
 			});
-			// socket.on("end", this.on_disconnected.bind(this));
 			socket.on("error", () => {
 				this.socket = null;
-				this.emit("disconnected");
 			});
 			socket.on("close", () => {
 				this.socket = null;


README.md
```diff
@@ -220,3 +219,8 @@ Godot has some Editor Settings that can help you if your workflow involves chang
 - **Editor Settings > Text Editor > Behavior > Files > Auto Reload Scripts on External Change**
 - **Editor Settings > Interface > Editor > Save on Focus Loss**
 - **Editor Settings > Interface > Editor > Import Resources When Unfocused**
+
+### I'm using Linux and the drag + shift drop isn't working. Why?
+
+Most likely you're using Wayland as display server, and there's a limitation of support in VSCode. Try
+running `code` through the terminal in the project directory with the flag `--ozone-platform=x11`.
```
Calinou:
```suggestion
Most likely you're using Wayland as display server, and there's a limitation of support in VS Code. Try
running `code` through the terminal in the project directory with the flag `--ozone-platform=x11`.
This will force VS Code to run through XWayland, where it works.
```

Calinou (approved):
Thanks!

PR #990, opened 2026-03-07
gdscript: allow color picking

paul-marechal:
Implement `DocumentColorProvider` by extracting Color expressions, interpreting them, and replacing arguments in place when possible.

Add the following configurations:

- `godotTools.colorPicker.precision`: Controls how many decimals to include when applying edits.
- `godotTools.colorPicker.padValues`: Whether or not to pad decimal numbers.
- `godotTools.colorPicker.uppercaseHex`: Whether or not to write hexadecimals in uppercase when applying edits.

It should now be possible to get a preview of the color expressed by common Color methods, as well as pick and choose from a color palette provided by VS Code.

Only static expressions are evaluated, if you use a variable as one of the parameters then it will be ignored.

Color constant expressions are considered "readonly": The palette will show but provide no edits.

Test file:

```gd
Color.CYAN

Color()

Color("YELLOW")
Color("#ffff00")
Color("YELLOW", 0.5)
Color("#ffff00", 0.5)

Color(0, 1, 0)
Color(0, 1, 0, 0.5)
Color(0, 1, 0, 0.2)

Color8(255, 0, 0)
Color8(255, 0, 0, 128)
Color8(255, 0, 0, 0)

Color.from_rgba8(
    255, # red
            255, #green
    255,) #blue

Color.hex(
    0xFF00FFFF # best color
)
```

Illustration:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -375,6 +375,21 @@
 					"type": "boolean",
 					"default": true,
 					"description": "Whether to enable inlay hints in GDResource (.tscn, .tres, etc) files"
+				},
+				"godotTools.colorPicker.precision": {
+					"type": "integer",
+					"default": 3,
+					"description": "Amount of decimals to include when using the color picker. 0 for arbitrary precision."
+				},
+				"godotTools.colorPicker.padValues": {
+					"type": "boolean",
+					"default": false,
+					"description": "Write either 1.2 (unpadded) or either 1.200 (padded)."
+				},
+				"godotTools.colorPicker.uppercaseHex": {
+					"type": "boolean",
+					"default": false,
+					"description": "Write either 0x008b8bff or either 0x008B8BFF."
 				}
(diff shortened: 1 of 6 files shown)


paul-marechal:
Ideally this information should come from Godot itself through LSP?

I had a lot of pain trying to implement an ad-hoc way of parsing expressions to allow editing arguments in place as I felt like this was a desirable behavior (being a user of the extension myself)... But honestly the code required to allow for this is a bit much so I'd understand if we'd want to scrap it.

DaelonSuzuka:
This looks good! I had starting building this feature last year but I got bogged down in supporting all the color formats and never finished it.

Give me a chance to play with this but I like what I see.

paul-marechal:
Cleaned up a little and rebased on master.

PR #993, opened 2026-03-09
Fix wrong inlay hint positions in large documents

### LINKED ISSUE #980: Undesirable inlay type hints displayed in incorrect positions in some scripts
### Godot version

4.6.1-stable

### VS Code version

1.109.5

2.6.1

### System information

Windows 11

### Issue description

When the experimental Inlay Hints for GDScript feature is enabled, in some scripts, unnecessary inlay type hints are displayed in nonsensical positions. The position changes depending on the document scroll position. The hints seem like they display the correct type if they're placed in a location where a type would appear, but may displayed even if there's already an explicit type annotation there. Other times, the hint is put in an invalid position, even if the middle of a word, with an "unknown" type. The faulty inlay hints are often displayed instead of legitimate hints that would be expected to appear elsewhere.

Image

Image

The exact circumstances this occurs are tricky to narrow down precisely, but I have reproducible script below. In this case, most of the stray type hints appear in the parameter list of a function definition. The following conditions seem to be needed to reproduce the bug:
* A certain number of leading lines in the script. May be just empty comments. The example script needs ~100 leading lines to reproduce the issue.
* Some following code where inlay type hints would be expected to appear, **even if it's commented out**. Changing the type inference assignments in the example to untyped assignments or adding explicit type annotations resolves the issue.

### Steps to reproduce

Add the following script in any project and scroll down so that `func foo(` is near the top of the editor window. Wait a moment for inlay hints to appear. Scroll around line-by-line to see the hints jump around to different spots.

<details>
<summary>Repro script</summary>

```gdscript
# # ...
# # ...
# # ...
# # ...
# # ...

# # ...
# # ...
# # ...
# # ...

# # ...
# # ...
# # ...

# # ...
# # ...
# # ...
# # ...
# # ...
# # ...
# # ...

# # ...
# # ...
# # ...
# # ...
# # ...
# # ...
# # ...

# # ...
# # ...
# # ...
# # ...
# # ...
# # ...
# # ...
# # ...
# # ...
# # ...
# # ...
# # ...
# # ...

# # ...
# # ...

# # ...
# # ...
# # ...

# # ...

# # ...
# # ...
# # ...

# # ...
# # ...
# # ...
# # ...
# # ...
# # ...
# # ...

# # ...
# # ...
# # ...
# # ...
# # ...
# # ...

# # ...

# # ...
# # ...
# # ...
# # ...
# # ...
# # ...

# # ...

# # ...
# # ...
# # ...
# # ...
# # ...
# # ...

# # ...
# # ...
# # ...
# # ...
# # ...
# # ...
# # ...
# # ...
# # ...
# # ...
# # ...
# # ...

# # ...
# # ...

func foo(
		_a: String,
		_b: String,
		_c: int,
		_d: String,
		_e: String,
		_f: bool,
		_g: int,
		_h: Array[String],
) -> void:
	pass

# func bar() -> String:
# 	var a := 0
# 	var b := 0
# 	var c := 0
# 	var d := 0
# 	var e := 0
# 	var f := 0
# 	var g := 0
# 	var h := 0
# 	var i := 0
# 	var j := 0
# 	var k := 0
# 	var l := 0
# 	var m := 0
# 	var n := 0
# 	var o := 0
# 	var p := 0

```
</details>

Nallebeorn:
Closes #980.

Fixes missing or misplaced inlay hints in GDScript and scene files, when scrolled down in a somewhat large document,

The extension respected the `range` parameter of `provideInlayHints` by extracting only that range substring for processing, but it didn't take the range into account when calculating positions for the LSP request or hint placement. So the position was incorrect any time the range start was non-zero.

The range is relative to vertical scroll position but with some padding. From my quick scan of the [relevant function in the VS Code source](https://github.com/microsoft/vscode/blob/072e99c1215d4acf06da82f9475d5a9b2fbc6e15/src/vs/editor/contrib/inlayHints/browser/inlayHintsController.ts#L460) the document size needed to reproduce should be something like two viewport heights tall plus 30 lines, so the bug is probably dependent on window and font size too :)

The linked issue has a repro GDScript (at least for me when fitting ~60 lines in the viewport). The bug also exists in scene files -- if you have a decent number of nodes with resources in the tree you'll find that the resource path hints are missing when scrolled down. I've attached a MRP with both a script and scene for testing convenience: [inlay-hints-test_2026-03-09_20-31-25.zip](https://github.com/user-attachments/files/25851822/inlay-hints-test_2026-03-09_20-31-25.zip)

### DIFF
--- a/src/providers/inlay_hints.ts
+++ b/src/providers/inlay_hints.ts
@@ -98,6 +98,7 @@ export class GDInlayHintsProvider implements InlayHintsProvider {
 	async provideInlayHints(document: TextDocument, range: Range, token: CancellationToken): Promise<InlayHint[]> {
 		const hints: InlayHint[] = [];
 		const text = document.getText(range);
+		const textStartOffset = document.offsetAt(range.start);
 		log.debug("Inlay Hints: provideInlayHints");
 
 		if (document.fileName.endsWith(".gd")) {
@@ -137,7 +138,7 @@ export class GDInlayHintsProvider implements InlayHintsProvider {
 				}
 				// TODO: until godot supports nested document symbols, we need to send
 				// a hover request for each variable declaration that is nested
-				const start = document.positionAt(match.index + match[0].length - 1);
+				const start = document.positionAt(textStartOffset + match.index + match[0].length - 1);
 
 				if (hasDetail) {
 					const symbol = symbols.find((s) => s.name === match[3]);
@@ -148,7 +149,7 @@ export class GDInlayHintsProvider implements InlayHintsProvider {
 					}
 				}
 
-				const hoverPosition = document.positionAt(match.index + match[1].length);
+				const hoverPosition = document.positionAt(textStartOffset + match.index + match[1].length);
 				const detail = await addByHover(document, hoverPosition);
 				if (detail) {
 					const hint = this.buildHint(start, detail);
@@ -166,7 +167,7 @@ export class GDInlayHintsProvider implements InlayHintsProvider {
 
 		for (const match of text.matchAll(/ExtResource\(\s?"?(\w+)\s?"?\)/g)) {
 			const id = match[1];
-			const end = document.positionAt(match.index + match[0].length);
+			const end = document.positionAt(textStartOffset + match.index + match[0].length);
 			const resource = scene.externalResources.get(id);
 
 			const label = `${resource.type}: "${resource.path}"`;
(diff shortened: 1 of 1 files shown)


DaelonSuzuka:
I've definitely seen the inlay hints positioned correctly in large documents scrolled 2+ pages down. I wonder why it's been intermittent for me...

Regardless, very simple fix, thank you for investigating!

PR #995, opened 2026-03-13
Update README.md to add instructions for XWayland workaround

andreygrigol:
I discovered recently that you can put the --ozone-platform=x11 snippet in the Exec Flags, and it'll also force to run VS Code through XWayland, without having to run it through the terminal. This facilitates a lot, so the user don't have to manually navigate to the folder of the project and running code with the flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,6 +223,7 @@ Godot has some Editor Settings that can help you if your workflow involves chang
 
 ### I'm using Linux and the drag + shift drop isn't working. Why?
 
-Most likely you're using Wayland as display server, and there's a limitation of support in VS Code. Try
-running `code` through the terminal in the project directory with the flag `--ozone-platform=x11`.
-This will force VS Code to run through XWayland, where it works.
+Most likely you're using Wayland as display server, and there's a limitation of support in VS Code. 
+In the exec flags, modify the snippet from `{project} --goto {file}:{line}:{col}` to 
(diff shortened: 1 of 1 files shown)


DaelonSuzuka:
I barely know anything about x11 vs wayland issues, is there anything else that this extension should be doing to support these differences?

Is anybody having problem launching Godot debug sessions from VSCode on wayland? Are there any flags there that would improve support?

I suppose this might be irrelevant if *Godot* runs properly in both environments?

(thanks for entertaining my stupid questions)

andreygrigol:
There's a problem with the plugin in Linux, where the Shift + Drag feature that plainly doesn't work when you're using Wayland as display server. I found a solution that running VS Code as a XWayland fixes this issue. I still need to study more about why this happens because, being honest w you, I have no clue why this happens. 

I added this information at the bottom of the README a few weeks ago, but you'd have to go manually to your directory on your system through the terminal and run VS Code with the X11 flag. But today I tried modifying the Exec Flags by adding the X11 parameter and it actually worked, a way more convenient form to "fix" the issue. 

In resume, the problem is not with Godot itself, it's a very specific problem of the plugin in VS Code when running on wayland. I'm just updating the README part about this, because there was already someone that ran into this problem, besides me lol

DaelonSuzuka:
Cool, thanks for sharing your research!

PR #1000, opened 2026-03-26
GDScript: Add abstract functions to highlight grammar

### LINKED ISSUE #834: Syntax Highlighting issue in `is not` operator
### Godot version

4.4 stable

### VS Code version

1.98.2

2.5.1

### System information

Mac os(Darwin arm64 24.3.0)

### Issue description

When using the `is not` operator for type checking in GDScript,
the syntax highlighting colors `is` and `not` differently, 
even though `is not` should be treated as a single operator according to the official documentation.

### Steps to reproduce

in godot editer
Image

in vscode
Image

```GDscript
# in GDscript
if node is not T:
    pass

x is Node
x is not Node
```

```python
# in python
if node is not T:
    pass

x is Node
x is not Node
```

### LINKED ISSUE #962: `:` character is not correctly escaped in a string when following an abstract function.
### Godot version

4.5.1 and 4.6.rc2

### VS Code version

1.108.2

2.5.1

### System information

Windows 11 x64

### Issue description

The `:` character is not correctly escaped in a string. When it's followed by a abstract function (who does not have an ending `:`.

Inside Godot Script Editor: (it's ok)
Image

There is the same issue in the VScode extension and in Github preview.

## Not working version

Image

```gdscript
@abstract
func _get_display_name() -> String

func _get_property_fullname(property: StringName) -> StringName:
	return StringName("%s:%s" % [_get_instance_name(), property])

func messed_up_func() -> void:
	prints("Everything here is broken")
```

## Working version
```gdscript
func _get_property_fullname(property: StringName) -> StringName:
	return StringName("%s:%s" % [_get_instance_name(), property])

func messed_up_func() -> void:
	prints("Everything here is OK")
```

About the Github broken part I don't know where to report it if anybody can point me a direction.

### Steps to reproduce

Create a script with the following code:
```gdscript
@abstract
func _get_display_name() -> String

func _get_property_fullname(property: StringName) -> StringName:
	return StringName("%s:%s" % [_get_instance_name(), property])

func messed_up_func() -> void:
	prints("Everything here is green")
```

HolonProduction:
Fixes: https://github.com/godotengine/godot-vscode-plugin/issues/962

Also adds `is not` (although not sure on the role, but it looks fine with the default color scheme)
Fixes: https://github.com/godotengine/godot-vscode-plugin/issues/834

### DIFF
--- a/syntaxes/GDScript.tmLanguage.json
+++ b/syntaxes/GDScript.tmLanguage.json
@@ -366,10 +366,11 @@
 			}
 		},
 		"class_is": {
-			"match": "\\s+(is)\\s+([a-zA-Z_]\\w*)",
+			"match": "\\s+(is)\\s+(not?)\\s+\\s+([a-zA-Z_]\\w*)",
 			"captures": {
 				"1": { "name": "storage.type.is.gdscript" },
-				"2": { "name": "entity.name.type.class.gdscript" }
+				"2": { "name": "storage.type.not.gdscript" },
+				"3": { "name": "entity.name.type.class.gdscript" }
 			}
 		},
 		"class_enum": {
@@ -507,7 +508,7 @@
 				"1": { "name": "keyword.language.gdscript storage.type.function.gdscript" },
 				"2": { "name": "entity.name.function.gdscript" }
 			},
-			"end": "(:)",
+			"end": "(:)|\n",
 			"endCaptures": { "1": { "name": "punctuation.section.function.begin.gdscript" } },
 			"patterns": [
 				{ "include": "#parameters" },


DaelonSuzuka:
@HolonProduction FYI I haven't been particularly strict on the token roles, there's a few contributors that come and correct me every once in a while.

The textmate tokenizer IS used for the formatter though (don't ask), so if you do something too outrageous with the highlighting the formatter tests should tell you. (this also means that when there's new grammar in GDScript, it's doubly helpful to make sure it's represented in the tests, because that checks both the formatter and the highlighting [indirectly])

PR #1006, opened 2026-04-14
Godot eval support

MichaelXt:
* calls Godot server evaluate on eval requests

* async version for verify_godot_version


DaelonSuzuka:
Fantastic, thanks! I poked at this once or twice but I didn't have enough context (or time) to get it working.

PR #1052, opened 2026-07-31
Add optional Godot editor terminal reuse

bcolloran:
This PR adds an option to reuse the existing Godot editor terminal rather than spawn a new one every time. I tend to move the terminal window out of the normal terminal space and into a tab in the code editing space, and it was killing me that my window arrangement would get messed up every time I needed to restart the editor for some reason.

Full disclosure, this was vibe coded by Codex 5.6 Sol. Absolutely zero hard feelings if you are auto closing vibe-coded PRs, I totally get it. But I thought I would go ahead and submit it because it looked like a pretty simple change, and I have found it to be a nice quality of life improvement, so I thought I would offer it up in case it would be of help to other folks, and since it didn't look too invasive or too sloppy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,10 @@ You can use the following settings to configure Godot Tools:
 
 The path to the Godot editor executable. _Under Mac OS, this is the executable inside of Godot.app._
 
+- `godotTools.editor.reuseTerminal`
+
+Reuse the existing Godot Editor terminal when relaunching the editor, preserving its location and layout.
+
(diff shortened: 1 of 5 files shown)


atirutw:
> this was vibe coded by Codex 5.6 Sol

I'm the type to judge based on code quality more than LLM usage, but my past experiences would suggest it's best to check whether the resulting codes is doing things the idiomatic or the intended way. Lots of LLMs fail at that.

Final decision is most likely up to Daelon though.

